PR #1, opened 2025-02-13
rebasing from latest rancher/rancher

slickwarren:
we need to be up-to-date with rancher/rancher -> main in order to start comparing latest tests and commits/PRs.


rancher-max (approved):
I didn't fully look at this, but looks like it covers the areas needed so LGTM

PR #4, opened 2025-02-13
create go package from actions. Move non-rancher actions to new folder

slickwarren:
This is a solution around bloating go.mod for anything that uses actions outside of this repo (i.e. rancher/rancher).

tldr; external repos should import rancher/tests/actions for say, rancher/rancher/tests/v2/integrations

This avoids importing non-rancher stuff into rancher i.e. harvester, tfp, etc.


rancher-max (approved):
I like this from a design perspective -- the rancher/tests/actions package should have requirements that are "baseline rancher" by nature, and the general rancher/tests package can have everything else. 

The goal here is clear: "do not bloat go.mod when standard actions from this repo are required in others". It seems like the distinguishing point is "if an action requires an update to go.mod from a rancher product then it belongs outside of the main actions folder." I'm not totally sure about the "interoperableactions" name, but everything else I think is fine.

rancher-max (approved):
It's not super clear if the full actions/codecoverage folder is being removed, or just the file inside it (thanks GH UI PR view)... if it's not the full folder, this should probably be updated to remove it. Otherwise LGTM

PR #6, opened 2025-02-19
deprecation guide 1.0

slickwarren:
Please follow the readme, and lmk any improvements we may need. 
This change will require we add tags to test runs, i.e. to run restrictedadmin tests we need to include one of 2.8, 2.9 or 2.10 tags in jenkins(and locally).


README.md
```diff
@@ -1,3 +1,76 @@
...
+
+##### Deprecating a subset of tests within a single test file
+1. create a new file
+2. move all deprecated tests to the new file
+3. rename the new file's suite, appropriate for the deprecated tests
+4. add go:build tags to the existing ones at the top of the new file
+
+#### Deprecating Actions
+1. if the **tests** being deprecated are spread across multiple packages, the deprecated action(s) used by said tests should be moved to a new file in the same folder of actions, named appropriately, signifying they contain deprecated actions
+2. If the **tests** being deprecated are iscolated to one package, the deprecated action(s) used by said tests should be moved to be a test helper within the deprecated test folder
```
anupama2501, line 52: typo: isolated

README.md
```diff
@@ -1,3 +1,76 @@
...
+
+## Branching Strategy
+
+main - active development branch
+
+stable - rebased from main after each rancher/rancher release. This branch should be used when importing this repo
+
+## Deprecation of a Feature for an Upcoming Release
+
+we use //go:build tags in our tests suites. The new branching strategy requires us to introduce a way to deprecate tests as well. We will be using go:build tags to deprecate tests.
```
anupama2501, line 13: `test suites` instead of tests suites.

validation/rbac/deprecated_restrictedadmin_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+		if newUser != nil {
+			rb.sequentialTestRBAC(tt.role, tt.member, newUser)
+			subSession := rb.session.NewSession()
+			defer subSession.Cleanup()
+		}
+	}
+}
+
+func (rb *RBTestSuite) TestRBACDynamicInput() {
```
anupama2501, line 135: I think we can remove the dynamic input here.. this was only introduced to validate any enhancements but since its deprecated we can only have the static tests for this.

validation/rbac/clusterandprojectroles/restrictedadmin_role_test.go
anupama2501, line 1: can we rename this to deprecated_restrictedadmin_role_test.go as well?

validation/rbac/deprecated_restrictedadmin_test.go
```diff
@@ -0,0 +1,169 @@
...
+func (rb *RBTestSuite) TestRBAC() {
+	tests := []struct {
+		name   string
+		role   rbac.Role
+		member string
+	}{
+		{"Restricted Admin", restrictedAdmin, restrictedAdmin.String()},
+	}
+
+	for _, tt := range tests {
```
anupama2501, line 118: nit, can we remove the for loop here as this test is for a single role.

README.md
```diff
@@ -1,3 +1,76 @@
...
 
-Welcome to the rancher test repo. 
\ No newline at end of file
+Welcome to the rancher test repo. 
+
+## Branching Strategy
+
+main - active development branch
+
+stable - rebased from main after each rancher/rancher release. This branch should be used when importing this repo
```
floatingman, line 9: What if we used a trunk-based strategy? I don't know what having two branches would solve since these are tests we should know right away if something is broken before merging into the main branch.

validation/rbac/clusterandprojectroles/restrictedadmin_role_test.go
```diff
@@ -0,0 +1,125 @@
+//go:build 2.8 || 2.9 || 2.10
```
floatingman, line 1: This can be !2.7 if we don't want to run it on version 2.7

validation/rbac/psa/deprecated_restrictedadmin_test.go
```diff
@@ -0,0 +1,249 @@
+//go:build (validation || infra.any || cluster.any || stress) && !sanity && !extended && (2.8 || 2.9 || 2.10)
```
floatingman, line 1: If we don't want to run this on 2.7, it should be `!2.7` to avoid changing it when Rancher versions change.

floatingman (changes requested):

README.md
```diff
@@ -1,3 +1,76 @@
...
+simply add go:build tags to the existing ones at the top of each file. 
+
+
+##### Deprecating a subset of tests within a single test file
+1. create a new file
+2. move all deprecated tests to the new file
+3. rename the new file's suite, appropriate for the deprecated tests
+4. add go:build tags to the existing ones at the top of the new file
+
+#### Deprecating Actions
```
floatingman, line 50: I would suggest adding a `//Deprecated:` comment to the functions so it shows up as deprecated in autocomplete in IDEs

validation/rbac/deprecated_restrictedadmin_test.go
anupama2501: Even though its for demonstration, it will still be merged and used in recurring runs. Since the test will be part of the recurring runs, I think it makes sense to modify it in this PR itself.

validation/rbac/deprecated_restrictedadmin_test.go
```diff
@@ -0,0 +1,169 @@
+//go:build (validation || infra.any || cluster.any || sanity) && !stress && !extended && (2.8 || 2.9 || 2.10)
```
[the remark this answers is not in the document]
anupama2501: but in that case, we would need to update this with each new release, e.g., !2.11, then later !2.12 etc.

README.md
slickwarren: briefly discussed offline, may reconsider if this is a problem in the future.

anupama2501 (commented):
lgtm. looks like the golang lint is running on the test helpers, but these are all used in the tests. probably want to disable it?

```
 Error: validation/rbac/deprecated_restrictedadmin.go:97:6: func `verifyRAUserCanCreateNamespace` is unused (unused)
  func verifyRAUserCanCreateNamespace(t *testing.T, standardClient *rancher.Client, project *management.Project, role rbacaction.Role) {}
```

anupama2501 (approved):

PR #19, opened 2025-02-21
adding missed RA deprecation test

slickwarren:
pending one discussion point on https://github.com/rancher/qa-tasks/issues/1695#issuecomment-2673030769

### DIFF
--- a/validation/fleet/snapshot/snapshot_restore_existing_gitrepo_test.go
+++ b/validation/fleet/snapshot/snapshot_restore_existing_gitrepo_test.go
@@ -35,11 +35,10 @@ const (
 
 type FleetWithSnapshotTestSuite struct {
 	suite.Suite
-	client         *rancher.Client
-	session        *session.Session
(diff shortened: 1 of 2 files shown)


validation/fleet/snapshot/snapshot_restore_existing_gitrepo_test.go
```diff
@@ -35,11 +35,10 @@ const (
 
 type FleetWithSnapshotTestSuite struct {
 	suite.Suite
-	client         *rancher.Client
-	session        *session.Session
-	fleetGitRepo   *v1alpha1.GitRepo
-	clustersConfig *etcdsnapshot.Config
```
anupama2501, line 41: clustersConfig was removed from fleet file. is that intentional?

slickwarren: yeah, it isn't used. Thanks golangci-lint!

anupama2501 (approved):

PR #21, opened 2025-02-21
adding pit to review modification to actions package

slickwarren:
also making sure fork'd contributions work the same way.


CODEOWNERS
```diff
@@ -1,4 +1,5 @@
 /actions/fleet @rancher/qa-pit-crew
+/actions/go.* @rancher/qa-pit-crew
```
rancher-max, line 2: Can multiple teams be added here? It might be good to also add the rancher/raft-core team here also.

lscalabrini01 (approved):

PR #22, opened 2025-02-24
ignore any potential ssh keys

slickwarren:


.gitignore
```diff
@@ -1 +1,2 @@
-.vscode/
\ No newline at end of file
+.vscode/
+*.pem
```
rancher-max, line 2: ahh yes very important, thank you!

Since we're adding to gitignore, a couple others that might be helpful:

```
.DS_Store
.idea/
```

rancher-max (approved):

.gitignore
slickwarren: good ideas!

lscalabrini01 (approved):

PR #23, opened 2025-02-24
slim down root's go.* since it isn't directly dependent on rancher/rancher now

slickwarren:


slickwarren:
not really working out, closing

PR #24, opened 2025-02-24
Update stable with main branch

rancher-max:


slickwarren:
spoke offline about this but, the history here is weird. We will see if this changes / is easier with time

PR #30, opened 2025-02-27
vai node metrics tables

brudnak:
Adding test to run sql query against vai databases to check for node metrics tables


joesims22 (approved):
lgtm

PR #31, opened 2025-02-28
Update Resync interval from 60minutes to 1minute

Priyashetty17:
CATTLE_RESYNC_DEFAULT accepts values in minutes, but it was mistakenly assumed to be in seconds. Updating the resync interval from 60 minutes to 1 minute.


brudnak (approved):
lgtm

PR #34, opened 2025-02-28
<main> Public repo fleet deployment on HA downstream clusters after upgrade

lscalabrini01:
Follow the rancher/rancher PR https://github.com/rancher/rancher/pull/48800


rancher-max (approved):
LGTM -- No need to do "assignees" as we are all added to the "reviewers" field

PR #37, opened 2025-03-04
Add AMI and reboot helper

susesgartner:
Created an AMI and reboot helper for OS checks


markusewalker (commented):
LGTM, left blocking comment to update the go.mod so that is handled before merging. Also, the PR needs to be rebased.

markusewalker (approved):

Josh-Diamond (approved):

PR #43, opened 2025-03-11
Fix harvester cloud provider rke2

slickwarren:
addresses: https://github.com/rancher/qa-tasks/issues/1646

issues:
my approach uses the kubeconfig from the cloud credential with no modification. The UI hits k8s/clusters/<harvester_imported_cluster_id>/v1/harvester/kubeconfig with a custom post command to inject an SA into the kubeconfig. If we want to perfectly emulate that behavior, I believe that belongs in a shepherd update / helper to the client. But it could possibly be done using the dynamic client. Either way its enough work to be a separate PR.

Otherwise, I slightly refactored cloud provider into its own package. This work isn't fully refactored, but separating it now will help a full refactor in the future should we choose to do so. 

Lastly I added some verifies to the harvester LB creation which only slightly differ from AWS, but enough IMO to pull it out into its own function(s). 

one-line or similar changes were basically just naming, or stuff that golangci-lint found in the package from the one-line change. 

Hostbusters, please take a look specifically with the rework around cloud provider package being pulled from provisioning / actions/permutations / actions/clusters (or a full review if you want 🥇 )


actions/go.mod
```diff
@@ -218,3 +220,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.3 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/rancher/shepherd => github.com/slickwarren/shepherd v0.0.0-20250311213324-40f961cc94c2
```
markusewalker, line 224: Leaving blocking comment so this is addressed before merging.

go.mod
```diff
@@ -262,3 +262,5 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.18.1 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.3 // indirect
 )
+
+replace github.com/rancher/shepherd => github.com/slickwarren/shepherd v0.0.0-20250311213324-40f961cc94c2
```
markusewalker, line 266: Leaving blocking comment so this isn't merged with the fork.

validation/Jenkinsfile.harvester
```diff
@@ -104,7 +105,7 @@ node("harvester-vpn-1") {
 
                 dir("./validation/.ssh") {
                   def decodedRsa = new String(AWS_SSH_RSA_KEY.decodeBase64())
-                  writeFile file: JENKINS_RKE_VALIDATION, text: decodedRsa
+                  writeFile file: AWS_RSA_KEY_NAME, text: decodedRsa
```
markusewalker, line 108: Does changing this back work? I ask because when Windows clusters were not working with Jenkins, I specifically had to put `JENKINS_RKE_VALIDATION` along with creating the parameter in Jenkins for it to work. I know this for Harvester and vSphere below, but making sure this doesn't break Windows tests.

markusewalker (commented):
Left some comments, nothing crazy. Can you provide Jenkins runs for the affected tests? Specifically:
- AWS cloud provider
- vSphere cloud provider
- Kubernetes upgrade (just RKE2 should suffice)

validation/Jenkinsfile.harvester
slickwarren: yeah it was causing an issue with the harvester job. AWS_RSA_KEY_NAME is now in the secrets manager. The original jenkinsfile didn't have an entry for JENKINS_RKE_VALIDATION so it was failing since the var didn't exist.

markusewalker: Got it. Especially as release testing is next week, can you run a simple windows provisioning check before this gets merged?

slickwarren: I don't have a great way to run windows on harvester right now, setting up the image is complicated 😓 I can run it for the vsphere one though

markusewalker (approved):

validation/upgrade/kubernetes_wins_test.go
```diff
@@ -68,7 +69,7 @@ func (u *UpgradeWindowsKubernetesTestSuite) TestUpgradeWindowsKubernetes() {
 			}
 
 			testConfig := clusters.ConvertConfigToClusterConfig(&cluster.ProvisioningInput)
-			upgradeDownstreamCluster(&u.Suite, tt.name, tt.client, cluster.Name, testConfig, cluster, tt.nodeSelector)
```
anupama2501, line 71: why are we removing the nodeSelector from here? last I remembered was that removing this would cause the workloads to be scheduled on linux nodes as well for windows clusters.

anupama2501 (commented):
Have a question

validation/upgrade/kubernetes_wins_test.go
slickwarren: It wasn't being used in the function, and thus golangci-lint was complaining about it

anupama2501: ah ok, thanks!

anupama2501 (approved):

PR #45, opened 2025-03-12
adding check for forked repos of shepherd, rancher. Speed up verify-changes action via parallelism

slickwarren:
This is to protect from accidentally merging PRs that point to fork'd repos of rancher or shepherd (since we often point to these during testing)

please check the commit CI checks for examples. 
addresses: https://github.com/rancher/qa-tasks/issues/1723

### DIFF
--- a/.github/scripts/check-forked-repos.sh
+++ b/.github/scripts/check-forked-repos.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+if grep -q "github.com/.*rancher/shepherd =>" go.mod; then
+    echo "Error: Forked version of rancher/shepherd detected in go.mod"
+    exit 1
+fi
+if grep -q "github.com/.*rancher/shepherd =>" actions/go.mod; then
+    echo "Error: Forked version of rancher/shepherd detected in actions/go.mod"
+    exit 1
+fi
+
+if grep -q "github.com/.*rancher/rancher =>" go.mod; then
+    echo "Error: Forked version of rancher/rancher detected in go.mod"
+    exit 1
+fi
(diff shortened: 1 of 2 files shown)


slickwarren:
link to failing check-forks verify: https://github.com/rancher/tests/actions/runs/13821711500

.github/scripts/check-forked-repos.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+if grep -q "github.com/.*rancher/shepherd =>" go.mod; then
```
anupama2501, line 5: guessing we are capturing the replace directive, shouldn't the pattern just be `github.com/rancher/shepherd =>`

.github/workflows/ci.yml
```diff
@@ -31,17 +44,27 @@ jobs:
...
           export PATH=$PATH:/home/runner/go/bin/
-      
+
       - name: Verify Go Mod
         run: ./.github/scripts/check-gomod.sh
 
       - name: Build Packages
         run: ./.github/scripts/go-build.sh
 
+  verify-golangci-lint:
```
anupama2501, line 58: I like how this is separated out. easier to debug when something fails.

.github/scripts/check-forked-repos.sh
slickwarren: sure I can rm the * check

anupama2501:
verify-golangci-lint has some failures. something to do with  `--new-from-rev` I think?

slickwarren:
it timed out, bumping the timer to 5 mins for that section

anupama2501 (approved):

rancher-max (approved):

PR #49, opened 2025-03-17
Update Ingress path from Exact to ImplementationSpecific

markusewalker:
### PR Description
In the snapshot tests, occassionally there have been reports of the following error message: `ingress contains invalid paths: path /index.html cannot be used with pathType Exact`. Looking closer into it, we should be using `ImplementationSpecific` as the path and not `Exact`.


Josh-Diamond (approved):

navsalehi (commented):
Aproved

PR #54, opened 2025-03-20
toggle secret filter test cases

brudnak:
vai can now support filtering by arbitrary labels so switched those cases from false, to true. another change in vai is exact matching, updated some tests to reflect this


Priyashetty17 (approved):

joesims22 (approved):
lgtm

PR #55, opened 2025-03-21
Test-a-Bot init commit

slickwarren:
addresses this exploratory issue: https://github.com/rancher/qa-tasks/issues/1754

Some notes for future works:
* doesn't look at structs / consts?
* doesn't work for non-go files i.e. yaml
* not tested on removed or added functions, probably won't work well with the current lookups
* WILL run every time there's a new push to the PR

[click here to see an example](https://github.com/slickwarren/tests/pull/2#issuecomment-2744195337) of what the comment looks like when modifying actions/clusters/clusters.go

notes about the file organization:
* creating a new file that is `workflow_run`, triggered when the `Verify Changes` action is received in Github
  * this gets around some permissions issues when PRs are opened from forks, as secrets aren't shared between the parent repo and any forks for security purposes.

### DIFF
--- a/.github/scripts/check-changed-tests.sh
+++ b/.github/scripts/check-changed-tests.sh
@@ -14,14 +14,14 @@ trap error_handler ERR
 TARGET_BRANCH="main"
 TEMP_DIR=$(mktemp -d)
 
-touch $TEMP_DIR/diff.used-anywhere
-touch $TEMP_DIR/diff.test-functions
+touch "$TEMP_DIR/diff.used-anywhere"
+touch "$TEMP_DIR/diff.test-functions"
 
 # returns the name of a function, supplied as an argument
 grep-for-function-name() {
     echo "$1" | grep func | \
     sed -E 's/.*func[[:space:]]+([a-zA-Z0-9_]+).*/\1/' | \
-    grep -v "(" | \ 
+    grep -v "(" | \
     grep -v "//" # omit comments and functions that are part of a suite ()
 }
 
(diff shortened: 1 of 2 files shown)


.github/scripts/check-changed-tests.sh
floatingman, line 1: I recommend running this script through shellcheck. It will give some pointers on making it more robust.

floatingman (commented):
Overall, it looks good, just one suggestion.

floatingman (approved):

rancher-max (commented):
This looks like a good start based on the output, but I'm wondering if we can narrow the recommended tests a little bit? From the example PR, it [recommended](https://github.com/slickwarren/tests/pull/2#issuecomment-2744195337) running tests that touch RKE1, but the only function in cluster.go that was updated was `NewK3SRKE2ClusterConfig`, so it would be overkill in that case to run RKE1 tests as well

slickwarren:
@rancher-max for actions changed, its complicated. Yes, that is the only function changed, but since that is used in permutations.go for the RunTestPermutations func, it will report anything that uses that function too. For now I'm not sure if checking conditionals within functions would work very well for reporting which tests to run. But that function should be reworked at some point anyways.

rancher-max (approved):
> @rancher-max for actions changed, its complicated. Yes, that is the only function changed, but since that is used in permutations.go for the RunTestPermutations func, it will report anything that uses that function too. For now I'm not sure if checking conditionals within functions would work very well for reporting which tests to run. But that function should be reworked at some point anyways.

That's fair! There's some benefit in that case too -- it will be glaringly obvious which functions are being overworked 😄

lscalabrini01 (approved):

PR #56, opened 2025-03-21
Update helper to use get instead of list

Priyashetty17:
List won't work for cluster member so updated helper `WaitForProjectFinalizerToUpdate` to use get instead of list.
Also, added a new helper to get the cluster Wrangler context, which is useful for most tests relying on the Wrangler context.


joesims22 (approved):
lgtm

PR #57, opened 2025-03-24
RBAC tests for Workloads - Job

Priyashetty17:
**Issue**: https://github.com/rancher/qa-tasks/issues/1342

This is one of the several upcoming PRs aimed at migrating the RBAC workload tests from Python to Go. **This PR specifically includes tests for Job only.**


joesims22 (dismissed):
lgtm

actions/workloads/job/verify.go
```diff
@@ -0,0 +1,22 @@
...
+	}
+
+	podTemplate := CreateContainerAndPodTemplate()
+
+	_, err = CreateJob(client, clusterID, namespace.Name, podTemplate, true)
+	if err != nil {
+		return err
+	}
+
+	return err
```
brudnak, line 21: Just a small thing - at the end of VerifyCreateJob, maybe return nil instead of return err for readability (since at that point err would be nil anyway). Otherwise this all looks good! Similar return also seen at `VerifyCreateDaemonSet`

brudnak (approved):
lgtm!

joesims22 (approved):

PR #58, opened 2025-03-24
add key-pair cleanup for aws in custodian

slickwarren:
https://github.com/rancher/qa-tasks/issues/1758

note- this is hard to test as-is, will probably just test on-prod...

### DIFF
--- a/scripts/custodian/aws.yaml
+++ b/scripts/custodian/aws.yaml
@@ -266,3 +266,88 @@ policies:
         op: delete
   actions:
     - type: delete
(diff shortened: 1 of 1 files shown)


rancher-max (approved):
LGTM -- let's just make sure jenkins ssh keys have the proper tags so that they aren't deleted in case they haven't been used recently.

slickwarren:
just added the correct tags to *-validation key pairs

markusewalker (approved):

PR #69, opened 2025-04-01
[main] P0 tests for GH-44663 

joesims22:
Automate tests in https://github.com/rancher/qa-tasks/issues/1635

### DIFF
--- a/actions/rbac/rbac.go
+++ b/actions/rbac/rbac.go
@@ -377,6 +377,32 @@ func CreateClusterRoleTemplateBinding(client *rancher.Client, clusterID string, 
 	return crtb, nil
 }
 
+// GetClusterRoleTemplateBindingsForUser fetches clusterroletemplatebindings for a specific user
+func GetClusterRoleTemplateBindingsForUser(rancherClient *rancher.Client, userID string) (*v3.ClusterRoleTemplateBinding, error) {
+	var matchingCRTB *v3.ClusterRoleTemplateBinding
+	err := kwait.PollUntilContextTimeout(context.TODO(), defaults.FiveSecondTimeout, defaults.OneMinuteTimeout, false, func(ctx context.Context) (done bool, pollErr error) {
+		crtbList, err := rbacapi.ListClusterRoleTemplateBindings(rancherClient, metav1.ListOptions{})
+		if err != nil {
+			return false, err
+		}
+
+		for _, crtb := range crtbList.Items {
+			if crtb.UserName == userID {
+				matchingCRTB = &crtb
+				return true, nil
+			}
+		}
+
+		return false, nil
+	})
+
+	if err != nil {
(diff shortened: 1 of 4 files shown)


actions/rbac/rbac.go
```diff
@@ -380,6 +380,37 @@ func CreateClusterRoleTemplateBinding(client *rancher.Client, clusterID string,
...
+		for _, crtb := range crtbList.Items {
+			if crtb.UserName == userID {
+				matchingCRTB = &crtb
+				return true, nil
+			}
+		}
+
+		return false, nil
+	})
+	return matchingCRTB, err
```
Priyashetty17, line 401: could you instead do:
```
if err != nil {
	return nil, fmt.Errorf("error while polling for crtb: %w", err)
}

return matchingCRTB, nil
```

actions/rbac/rbac.go
```diff
@@ -380,6 +380,37 @@ func CreateClusterRoleTemplateBinding(client *rancher.Client, clusterID string,
...
+}
+
+// GetClusterRoleTemplateBindings fetches a specific clusterroletemplatebinding by name
+func GetClusterRoleTemplateBindingByName(client *rancher.Client, namespace, name string) (*v3.ClusterRoleTemplateBinding, error) {
```
Priyashetty17, line 405: Could you get ClusterRoleTemplateBinding directly in the test instead? Is this function necessary?

validation/rbac/crtb/crtb.go
```diff
@@ -0,0 +1,108 @@
...
+	TrueConditionStatus              metav1.ConditionStatus = "True"
+	FalseConditionStatus             metav1.ConditionStatus = "False"
+	CompletedSummary                                        = "Completed"
+	SubjectExists                                           = "SubjectExists"
+	LabelsReconciled                                        = "LabelsReconciled"
+	BindingExists                                           = "BindingExists"
+	CRTBLabelsUpdated                                       = "CRTBLabelsUpdated"
+	ClusterRolesExist                                       = "ClusterRolesExist"
+	ClusterRoleBindingsExists                               = "ClusterRoleBindingsExists"
+	ServiceAccountImpersonatorExists                        = "ServiceAccountImpersonatorExists"
```
Priyashetty17, line 25: Since these constants are private (only used within the same package), it's best practice to follow lowerCamelCase. As part of this PR, could you also update this in https://github.com/rancher/tests/blob/main/validation/rbac/grb/grb.go#L22-L31?

actions/rbac/rbac.go
joesims22: Good call, I'll remove the func here and get the CRTB directly in the test.

Priyashetty17 (approved):
lgtm

brudnak (approved):
lgtm! :shipit:

PR #94, opened 2025-04-21
Update GET prtb in the tests to use prtb namespace instead of project name

Priyashetty17:
A new field was added to projects to specify the namespace for storing PRTBs. So, the PRTB namespace is now a combination of the cluster name and the project name, rather than just the project name. GET PRTB method has been updated in this PR to use the PRTB namespace instead of the project name as the namespace to ensure backward compatibility.


joesims22 (commented):
lgtm

joesims22 (approved):
lgtm

PR #99, opened 2025-04-25
Automate tests for GH-47359

joesims22:
Automate tests in https://github.com/rancher/qa-tasks/issues/1702

### DIFF
--- a/validation/rbac/grb/grb.go
+++ b/validation/rbac/grb/grb.go
@@ -29,6 +29,10 @@ const (
 	namespacedRoleBindingReconciled                           = "NamespacedRoleBindingReconciled"
 	fleetWorkspacePermissionReconciled                        = "FleetWorkspacePermissionReconciled"
 	clusterAdminRoleExists                                    = "ClusterAdminRoleExists"
+	upnString                                                 = "testuser1"
+	principalDisplayNameAnnotation                            = "auth.cattle.io/principal-display-name"
+	testPrincipalDisplayName                                  = "testPrincipalDisplayName"
+	testGroupPrincipalName                                    = "testGroupPrincipalName"
 )
 
 var (
@@ -43,6 +47,15 @@ var (
 				Verbs:     []string{"*"},
 			},
 		},
+	}
+
+	customGlobalRoleBinding = v3.GlobalRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "",
+			Annotations: map[string]string{},
+		},
+		GlobalRoleName:    "",
+		UserPrincipalName: upnString,
 	}
 )
 
@@ -139,5 +152,21 @@ func verifyGlobalRoleBindingStatusField(grb *v3.GlobalRoleBinding, isAdminGlobal
 		}
 	}
 
+	return nil
+}
+
+func verifyUserByPrincipalIDExists(client *rancher.Client, principalID string) error {
+	userList, err := client.WranglerContext.Mgmt.User().List(metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("Failed to list users %w", err)
+	}
+
+	for _, user := range userList.Items {
+		for _, principalId := range user.PrincipalIDs {
+			if principalId == principalID {
+				return nil
+			}
+		}
+	}
 	return nil
 }
--- a/validation/rbac/grb/grb_upn_test.go
+++ b/validation/rbac/grb/grb_upn_test.go
@@ -1,0 +1,123 @@
+//go:build (validation || infra.any || cluster.any || extended) && !sanity && !stress && !(2.8 || 2.9 || 2.10)
+
+package grb
+
+import (
+	"testing"
+
+	"github.com/rancher/shepherd/clients/rancher"
+	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	"github.com/rancher/shepherd/extensions/clusters"
+	namegen "github.com/rancher/shepherd/pkg/namegenerator"
+	"github.com/rancher/shepherd/pkg/session"
+	rbacapi "github.com/rancher/tests/actions/kubeapi/rbac"
+	"github.com/rancher/tests/actions/rbac"
+	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+type GlobalRoleBindingUserPrincipalNameTestSuite struct {
+	suite.Suite
+	client  *rancher.Client
+	session *session.Session
+	cluster *management.Cluster
+}
+
+func (upn *GlobalRoleBindingUserPrincipalNameTestSuite) TearDownSuite() {
+	upn.session.Cleanup()
+}
+
+func (upn *GlobalRoleBindingUserPrincipalNameTestSuite) SetupSuite() {
+	upn.session = session.NewSession()
+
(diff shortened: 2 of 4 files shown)


validation/rbac/grb/grb.go
```diff
@@ -29,6 +29,9 @@ const (
 	namespacedRoleBindingReconciled                           = "NamespacedRoleBindingReconciled"
 	fleetWorkspacePermissionReconciled                        = "FleetWorkspacePermissionReconciled"
 	clusterAdminRoleExists                                    = "ClusterAdminRoleExists"
+	upnString                                                 = "activedirectory_user://CN=testuser1,CN=Users,DC=ad,DC=ians,DC=farm"
```
dasarinaidu, line 32: Why do we need a auth here? are we using activedirectory in this test? OR just a string? 
If it is a string better to use a general string than a auth reference.

validation/rbac/grb/grb.go
```diff
@@ -141,3 +155,19 @@ func verifyGlobalRoleBindingStatusField(grb *v3.GlobalRoleBinding, isAdminGlobal
 
 	return nil
 }
+
+func verifyUserInUserPrincipalNameIsCreated(client *rancher.Client, userPrincipalName string) error {
```
Priyashetty17, line 159: Instead of using List() and looping through all users, can you use WranglerContext.Mgmt.User().Get() with the specific user ID/name? This might be more efficient and avoids iterating over the entire user list.
```

func verifyUserExistsByNameAndPrincipalID(client *rancher.Client, userName, userPrincipalID string) error {
	user, err := client.WranglerContext.Mgmt.User().Get(userName, metav1.GetOptions{})
	if err != nil {
		return fmt.Errorf("failed to get user %q: %w", userName, err)
	}

	for _, principalID := range user.PrincipalIDs {
		if principalID == userPrincipalID {
			return nil
		}
	}

	return fmt.Errorf("user %q found, but principal ID %q not associated", userName, userPrincipalID)
}
```

validation/rbac/grb/grb_upn_test.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (upn *GlobalRoleBindingUserPrincipalNameTestSuite) TestCreateGlobalRoleBindingWithUserPrincipalName() {
+	subSession := upn.session.NewSession()
+	defer subSession.Cleanup()
+
+	log.Info("Create a global role that inherits cluster-member")
+	globalRoleWithInheritedClusterMember, _, err := createGlobalRoleAndUser(upn.client)
+	globalRoleWithInheritedClusterMember.InheritedClusterRoles = []string{"cluster-member"}
+	require.NoError(upn.T(), err)
```
Priyashetty17, line 51: This will not create a Global role that inherits cluster member. It only modifies the InheritedClusterRoles field locally in memory after the role has already been created. You need to do this instead:
```
log.Info("Create a global role that inherits cluster-member")
globalRoleWithInheritedClusterMember, _, err := createGlobalRoleAndUser(upn.client)
require.NoError(upn.T(), err)

globalRoleWithInheritedClusterMember.InheritedClusterRoles = []string{rbac.ClusterMember.String()}
_, err = rbacapi.UpdateGlobalRole(upn.client, &globalRoleWithInheritedClusterMember)
require.NoError(upn.T(), err)
```
Note: rbacapi "github.com/rancher/tests/actions/kubeapi/rbac"

This comment applies to all the tests in this file

validation/rbac/grb/grb_upn_test.go
```diff
@@ -0,0 +1,110 @@
+package grb
```
Priyashetty17, line 1: Build tags are missing, could you please add them? Since this functionality applies only to versions >=2.11, make sure the build tag excludes versions 2.8, 2.9, and 2.10 i.e., `&& !(2.8 || 2.9 || 2.10))`

validation/rbac/grb/grb.go
joesims22: We are not specifically using Active Directory in this test.

In the original [issue](https://github.com/rancher/rancher/issues/47359) the enhancement was for matching GRB behavior to CRTBs in regard to being able to pre-populate RBAC for an Active Directory user, however setting the `userPrincipalName` field to a generic string will create a user as well. 

Since we aren't actually enabling Active Directory in this automated testcase I agree the string can be changed to a general string.

validation/rbac/grb/grb.go
```diff
@@ -44,6 +47,17 @@ var (
 			},
 		},
 	}
+
+	customGlobalRoleBinding = v3.GlobalRoleBinding{
```
Priyashetty17, line 51: The annotation is not applicable to all tests, correct? Could you update this to accept annotations as a dynamic field, so it can be more generic and reusable by other tests that don’t necessarily use the principal name? Something like this:
```
customGlobalRoleBinding = v3.GlobalRoleBinding{
	ObjectMeta: metav1.ObjectMeta{
		Name:        "",
		Annotations: map[string]string{},
	},
	GlobalRoleName:    "",
	UserPrincipalName: upnString,
}

```
In the test that uses annotation, you can then set the annotations as needed, for example:
```
customGlobalRoleBinding.Annotations["auth.cattle.io/principal-display-name"] = testPrincipalDisplayName
_, err = rbacapi.CreateGlobalRoleBinding(upn.client, &customGlobalRoleBinding)
```
you could define "auth.cattle.io/principal-display-name" as a constant to avoid hardcoding it multiple times.

Priyashetty17 (changes requested):
Left a few comments

actions/users/users.go
```diff
@@ -20,4 +20,3 @@ func UpdateUserPassword(client *rancher.Client, user *management.User, password
 
 	return nil
 }
-
```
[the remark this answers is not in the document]
joesims22: Oops, good catch. This got deleted when I removed a helper func I didn't end up needing. Will re-add.

validation/rbac/grb/grb.go
joesims22: Good call, I will update the variable and test accordingly.

validation/rbac/grb/grb.go
joesims22: We spoke offline about this -- keeping previous logic but updating function name and param name for clarity.

validation/rbac/grb/grb_upn_test.go
```diff
@@ -0,0 +1,123 @@
...
+	require.NoError(upn.T(), err)
+
+	globalRoleWithInheritedClusterMember.InheritedClusterRoles = []string{rbac.ClusterMember.String()}
+	_, err = rbacapi.UpdateGlobalRole(upn.client, globalRoleWithInheritedClusterMember)
+	require.NoError(upn.T(), err)
+
+	log.Info("Create a globalrolebinding with userPrincipalName and principal-display-name annotation set")
+	customGlobalRoleBinding.Name = namegen.AppendRandomString("testgrb")
+	customGlobalRoleBinding.GlobalRoleName = globalRoleWithInheritedClusterMember.Name
+	customGlobalRoleBinding.Annotations["auth.cattle.io/principal-display-name"] = testPrincipalDisplayName
```
Priyashetty17, line 85: Replace the string "auth.cattle.io/principal-display-name" with constant `principalDisplayNameAnnotation`

Priyashetty17 (approved):
lgtm

Priyashetty17 (approved):

dasarinaidu (approved):

PR #106, opened 2025-05-05
updating azure to remove unused public IPs every 2 days.

slickwarren:
tested locally, correctly marks and removes all public IPs not tagged as 'locked' and not used currently. 

note the normal `tag` and `name` filters don't exist for this resource. Since these are protected resources, in the sense that if they are used elsewhere, then you can't remove them, not having these filters shouldn't limit us as the API won't delete them. i.e. 
```
Microsoft.Network/publicIPAddresses/maybeidk2-ip can not be deleted since it is still allocated to resource

```

### DIFF
--- a/scripts/custodian/azure.yaml
+++ b/scripts/custodian/azure.yaml
@@ -307,3 +307,41 @@ policies:
         op: delete
   actions:
     - type: delete
+
+# Public IP Address
+
+
+- name: az-mark-ip-for-deletion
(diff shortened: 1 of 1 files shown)


slickwarren:
supports https://github.com/rancher/qa-tasks/issues/1802

rancher-max (approved):

daviswill2 (approved):

PR #111, opened 2025-05-07
Providers rework

susesgartner:
Rework providers so that generic permutations can replace old permutations without disruption. This PR includes a rework to all rke2 tests to new permutations as well as a few QOL changes to defaults.


Josh-Diamond (dismissed):

markusewalker (approved):

validation/provisioning/rke2/provisioning_node_driver_test.go
```diff
@@ -37,24 +41,26 @@ func (r *RKE2NodeDriverProvisioningTestSuite) TearDownSuite() {
...
 
-		r.provisioningConfig.RKE2KubernetesVersions = rke2Versions
-	}
+	k8sPermutation, err := permutationdata.CreateK8sPermutation(r.client, "rke2", cattleConfig)
+	require.NoError(r.T(), err)
+
+	cniPermutation, err := permutationdata.CreateCNIPermutation(cattleConfig)
+	require.NoError(r.T(), err)
+
+	permutedConfigs, err := permutations.Permute([]permutations.Permutation{*k8sPermutation, *providerPermutation, *cniPermutation}, cattleConfig)
```
slickwarren, line 60: <discussion point for later, not blocking> I wish this was abstracted from the test. Since this will be done every provisioning test essentially, it is a little long to have to put in each test. It would be nice to have something, maybe an action, that knows all the permutations for provisioning and just does those. 

which actually brings me to another point, will each thing we want to permute on need to be explicit? i.e. anything not k8s, provider, or cni will not be permuted on when running this test. i.e. say I wanted to permute on my AWS AMIs, If I had a list of amis in that field and tried to run this test, it would just fail I assume? 

in other words, when I'm running the test, how can I know which variables I'm allowed to permute on (or will it just be whatever's in the config later, and this is just a first start?)

validation/provisioning/rke2/provisioning_node_driver_test.go
```diff
@@ -108,22 +114,33 @@ func (r *RKE2NodeDriverProvisioningTestSuite) TestProvisioningRKE2Cluster() {
...
 
-		permutations.RunTestPermutations(&r.Suite, tt.name, tt.client, &provisioningConfig, permutations.RKE2ProvisionCluster, nil, nil)
+				provider := provisioning.CreateProvider(clusterConfig.Provider)
+				credentialSpec := cloudcredentials.LoadCloudCredential(string(provider.Name))
+				machineConfigSpec := machinepools.LoadMachineConfigs(string(provider.Name))
+
+				clusterObject, err := provisioning.CreateProvisioningCluster(tt.client, provider, credentialSpec, clusterConfig, machineConfigSpec, nil)
+				require.NoError(r.T(), err)
+
+				provisioning.VerifyCluster(r.T(), tt.client, clusterConfig, clusterObject)
```
slickwarren, line 137: https://github.com/rancher/tests/blob/9310be29e1e9cadcc54b0fc5e20824eacf2dcffe/actions/provisioning/permutations/permutations.go#L156 is missing in tests that use new permutations. If the cloud provider was enabled, it should be verified.

slickwarren (changes requested):
some non-blocking comments below, but the main concern I have is running cloud provider tests in the changed tests that now run the new permutations 

Otherwise, the changes look good.

validation/provisioning/rke2/post_kdm_oob_release_test.go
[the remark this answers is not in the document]
susesgartner: Yes I checked with them. They are no longer using it.

actions/machinepools/machinepools.go
```diff
@@ -33,10 +35,83 @@ const (
...
+	AWSMachineConfigsKey           = "awsMachineConfigs"
+	AzureMachineConfigsKey         = "azureMachineConfigs"
+	DOMachineConfigsKey            = "doMachineConfigs"
+	HarvesterMachineConfigsKey     = "harvesterMachineConfigs"
+	LinodeMachineConfigsKey        = "linodeMachineConfigs"
+	VmwarevsphereMachineConfigsKey = "vmwarevsphereMachineConfigs"
 )
 
+// MachineConfigs is the main struct needed to create a machine pool depending on the outside cloud service provider
+type MachineConfigs struct {
```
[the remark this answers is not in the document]
susesgartner: This is something I was thinking about addressing at a later date. Cloud credentials utilizes a similar struct to load the credentials at the test level (this was the case before credential rework), this change was made so that we have parity between our various objects. Overall from an architecture standpoint I have been splitting our load functions from our create functions for modularity and since there was no consistency between how we loaded different objects like credentials, machinePools ... etc I just tried to make them consistent. I do think we need to take another pass at this and probably move these load functions into the provider (and delete these structs) but for now I think parity is probably fine.

validation/provisioning/rke2/provisioning_node_driver_test.go
susesgartner: I do agree that this can be abstracted, for now i'm thinking about leaving it and we can abstract that as a QOL later down the line.

susesgartner: You are correct, new permutations is explicit and AMI's would not be handled unless you created a permutation for it. I did mess around with the idea of making all fields permutable but decided against it for the time being due to a few concerns I had around permutation defaults i.e kubernetesVersion set to all or "". In addition to the defaults issue you also have to consider that some permutations are conditional on the provider i.e AMI is an aws field, this would make things very complicated if you tried making all fields permutable. 

Overall I just think its a lot easier and more clear to have things be explicit and if your wondering which fields your allowed to permute in any given test you can just look at the readme OR just look at the test itself and its pretty easy to tell.

validation/provisioning/rke2/provisioning_node_driver_test.go
slickwarren: one small nit here, discussed offline

slickwarren (approved):
If you have time for one more jenkins run for any of these, would be appreciated. not blocking, LGTM

validation/provisioning/rke2/provisioning_node_driver_test.go
slickwarren: spoke offline, this will be an ongoing effort with the new permutations. Resolving as this effort will be separate.

markusewalker (approved):

PR #115, opened 2025-05-15
Add qase test uploader

rancher-max:
This is stage 1 of updating the qase test reporting. This allows test definitions to live closer to the code and update to Qase whenever it is run. I still need to create a GH Action that runs on PR merge that runs this new code.

This will allow schemas to be created that define the test name, the suite it lives in, any description, and the test steps. All of this across any project in our Qase instance.


lscalabrini01 (dismissed):

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+	Automation  int32           `json:"automation,omitempty"`
+	Steps       []TestCaseSteps `json:"steps,omitempty"`
+}
+
+var (
+	qaseToken = os.Getenv(QaseTokenEnvVar)
+	Client    *qase.APIClient
+)
+
+// SetupQaseClient creates a new Qase client from the api token environment variable QASE_AUTOMATION_TOKEN
```
slickwarren, line 33: instead of being read in as an environment variable, can this be part of the RANCHER_CATTLE_CONFIG config.yaml and read in that way? Like we do for rancher objects and such.

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+		testCases = append(testCases, tempResult.Result.Entities...)
+		numOfTestCases = tempResult.Result.Count
+		offSetCount += 10
+	}
+
+	return testCases, nil
+}
+
+// GetTestSuite retrieves a Test Suite by name within a specified Qase Project if it exists
+func GetTestSuite(project, suite string) (qase.Suite, error) {
```
slickwarren, line 84: I believe this should be returning `*qase.Suite`

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+	Description string          `json:"description,omitempty"`
+	Title       string          `json:"title"`
+	SuiteId     int64           `json:"suite_id,omitempty"`
+	Automation  int32           `json:"automation,omitempty"`
+	Steps       []TestCaseSteps `json:"steps,omitempty"`
+}
+
+var (
+	qaseToken = os.Getenv(QaseTokenEnvVar)
+	Client    *qase.APIClient
```
slickwarren, line 30: typically, we don't use var for clients. Instead, you can make the client part of a struct and define all its functions accordingly (similar to the structure of test files) or just pass the client along as a parameter to each function.

i.e. 
```
type QaseService struct {
  Client *qase.Client
}

func (q *QaseService) Setup {
...
}

...
func (q *QaseService) GetTestCase() () {
...}
```

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+	logrus.Debugf("Getting test suite \"%s\" in project %s\n", suite, project)
+	localVarOptionals := &qase.SuitesApiGetSuitesOpts{
+		FiltersSearch: optional.NewString(suite),
+	}
+	qaseSuites, _, err := Client.SuitesApi.GetSuites(context.TODO(), project, localVarOptionals)
+	if err != nil {
+		return qase.Suite{}, err
+	}
+
+	for _, qaseSuite := range qaseSuites.Result.Entities {
```
slickwarren, line 94: same idea here as with the test cases for loop

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+		if qaseSuite.Title == suite {
+			return qaseSuite, nil
+		}
+	}
+	return qase.Suite{}, fmt.Errorf("test suite \"%s\" not found in project %s", suite, project)
+}
+
+// CreateTestSuite retrieves a Test Suite by name within a specified Qase Project if it exists
```
slickwarren, line 102: this description seems off. I would expect it to do this 
```
// CreateTestSuite creates a Test Suite, given a name, within a specified Qase Project if the suite does not already exist
```

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+
+func createTestCases(project string, testCase qase.TestCaseCreate) error {
+	_, _, err := Client.CasesApi.CreateCase(context.TODO(), testCase, project)
+	if err != nil {
+		return fmt.Errorf("failed to create test case: \"%s\". Error: %v", testCase.Title, err)
+	}
+	return nil
+}
+
+func updateTestCases(project string, testCase qase.TestCaseUpdate, id int32) error {
```
slickwarren, line 168: since this only updates one test case, rename to `updateTestCase`

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+			err = createTestCases(project, qaseTest)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func createTestCases(project string, testCase qase.TestCaseCreate) error {
```
slickwarren, line 160: since this only creates one test case, rename to `createTestCase`

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+	for _, tc := range testCases {
+		logrus.Info("Uploading test case:\n\tProject: ", project, "\n\tTitle: ", tc.Title, "\n\tDescription: ", tc.Description, "\n\tSuiteId: ", tc.SuiteId, "\n\tAutomation: ", tc.Automation, "\n\tSteps: ", tc.Steps)
+
+		existingCase, err := GetTestCase(project, tc.Title)
+		if err == nil {
+			var qaseTest qase.TestCaseUpdate
+			qaseTest.Title = tc.Title
+			qaseTest.Description = tc.Description
+			qaseTest.SuiteId = tc.SuiteId
+			qaseTest.Automation = tc.Automation
```
slickwarren, line 124: <nit> this looks reusable enough, I'm surprised upstream doesn't have a built in conversion for it? 

i.e. adding a function like this would be pretty imo https://github.com/rancher/tests/blob/main/actions/clusters/clusterconfig.go#L41

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+			for _, step := range tc.Steps {
+				var qaseSteps qase.TestCaseUpdateSteps
+				qaseSteps.Action = step.Action
+				qaseSteps.ExpectedResult = step.ExpectedResult
+				qaseSteps.Data = step.InputData
+				qaseSteps.Position = step.Position
+				qaseTest.Steps = append(qaseTest.Steps, qaseSteps)
```
slickwarren, line 131: same idea for qaseSteps as for qaseTest

slickwarren: in fact, this whole thing can probably be broken out into its own helper function since its shared between the if/else here and looks like it would generally be helpful on its own.

validation/pipeline/qase/qase.go
slickwarren: also, do these not get added to any particular test suite?

validation/pipeline/qase/qase.go
slickwarren: also, do these not get added to any particular test suite?

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/rancher/tests/validation/pipeline/qase"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	_, callerFilePath, _, _ = runtime.Caller(0)
+	basepath                = filepath.Join(filepath.Dir(callerFilePath), "..", "..", "..", "..")
```
slickwarren, line 17: would be more in favor of a const `../../../..` and calling the filepath.Join in functions instead of saving it as a var.

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+
+// getRawSchemaData retrieves the tests from schemas.md files defined within each Go package and returns a map of the string content within each suite
+func getRawSchemaData() (map[string][]string, error) {
+	schemaMap := make(map[string][]string)
+	var content []byte
+	err := filepath.Walk(basepath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if strings.Contains(info.Name(), "schemas.md") {
```
slickwarren, line 61: can you put `schemas.md` as a const

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+			content, err = os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+			c := string(content)
+			project := extractSubstrings(c, `# (.+) Schemas\n`)[0][1]
```
slickwarren, line 67: can you put all these regex strings as consts instead so that they have a short description of what they're doing

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+			if _, ok := schemaMap[project]; ok {
+				schemaMap[project] = append(schemaMap[project], strings.Split(c, "\n---\n")...)
+			} else {
```
slickwarren, line 70: same for this string, as it seems to be 'columns' or 'rows' string in the schema

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		err = qase.UploadTests(project, cases)
+		if err != nil {
+			logrus.Errorln("Error uploading tests:", err)
+		}
+	}
+}
+
+// getRawSchemaData retrieves the tests from schemas.md files defined within each Go package and returns a map of the string content within each suite
```
slickwarren, line 53: might be helpful to say what the expected contents of the map are, i.e. 
[projectName] : []{testCaseNames}

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+	schemaMap, err := getRawSchemaData()
+	if err != nil {
+		logrus.Error("Error retrieving test schemas: ", err)
+		return
+	}
+
+	for project, schemas := range schemaMap {
+		var cases []qase.TestCase
+		for _, schema := range schemas {
+			suite := extractSubstrings(schema, `## Test Suite: (.+)\n`)[0][1]
```
slickwarren, line 32: would be nice to document somewhere that `## Test Suite: ` is required in the schema

slickwarren: i.e. what if there's an error / typo in schemas.md, will this create false / erroneous suites?

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+			testSuite, suiteErr := qase.GetTestSuite(project, suite)
+			testSuiteId := testSuite.Id
```
slickwarren, line 34: does this not err if there's no existing suite yet?

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+			if suiteErr != nil {
+				testSuiteId, _ = qase.CreateTestSuite(project, suite)
```
slickwarren, line 36: we should capture this error

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+		testCase.Title = test[1]
+		testCase.Description = test[2]
+		testCase.Automation = 2
+		testCases = append(testCases, testCase)
+	}
+
+	index := 0
+	var testSteps qase.TestCaseSteps
+	for _, step := range steps {
+		stepPosition, err := strconv.ParseInt(step[1], 10, 32)
```
slickwarren, line 98: can you put these ints as consts so we know what they're for.

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+		if err != nil {
+			logrus.Error("Error converting string to int: ", err)
+			return nil, err
+		}
+		if stepPosition == 1 && testSteps != (qase.TestCaseSteps{}) {
```
slickwarren, line 103: so if no test step was added after looking at the first line in testSteps from the file, don't increment index. What is the consequence of this? seems that we would remain on the same test case until the steps aren't empty (looking at line 121)

validation/pipeline/qase/schemaupload/main.go
```diff
@@ -0,0 +1,143 @@
...
+		testSteps.Position = int32(stepPosition)
+		testSteps.Action = strings.TrimSpace(step[2])
+
+		if isFile(strings.TrimSpace(step[3])) {
+			fileContent, err := os.ReadFile(filepath.Join(basepath, strings.TrimSpace(step[3])))
+			if err != nil {
+				logrus.Error("Error reading file: ", err)
+				return nil, err
+			}
+			testSteps.InputData = strings.TrimSpace(step[3]) + "\n\n" + string(fileContent)
```
slickwarren, line 116: why include the filename in the test when reading contents from the file

slickwarren (changes requested):
sorry lots of comments. Mostly around moving stuff around for clarify on the code. 

if you haven't already, can you link in a private channel to the qase project where this is working?

validation/pipeline/qase/qase.go
rancher-max: I like using an environment variable here instead of a config as this is sensitive data and should just be part of the running "environment," so to speak.

validation/pipeline/qase/qase.go
rancher-max: They do! That's part of the TestCaseCreate data -- a `SuiteId` field.

validation/pipeline/qase/schemaupload/main.go
rancher-max: I need to get the root directory path a few times and this just made it easier instead of redoing the logic each time

validation/pipeline/qase/schemaupload/main.go
rancher-max: yeah I was trying to have the documentation with the pit_schemas.md example file that's in there. The overall structure of the file is very important to follow, and if it's not followed then it will definitely lead to weird errors... I tried to do MD file unmarshalling but things got very messy and I didn't like it.

validation/pipeline/qase/schemaupload/main.go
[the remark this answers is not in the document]
rancher-max: That's actually what I had originally! I changed it for two reasons:
1. Multiple teams may have schemas defined within the same folder and/or that use different Qase Project Codes, and I didn't have a great way to differentiate that.
2. The test steps may have data that require a .md file in some cases, so I wanted to leave that open and not just grab all .md files.

validation/pipeline/qase/schemaupload/main.go
rancher-max: This is just for the ParseInt function specifically -- string to parse, base 10, and must fit into a 32 bit integer (since it gets assigned into an int32 later)

validation/pipeline/qase/schemaupload/main.go
rancher-max: If there aren't any test steps, then it should never reach this loop, so it should be okay... but I didn't do any error handling for if they are very malformed.

validation/pipeline/qase/schemaupload/main.go
rancher-max: Depending on the file type and its data, uploading the string to qase doesn't always look great. So including the relative path to the file so that anyone following the steps directly from Qase can go back and grab the data directly there if need be.

validation/pipeline/qase/qase.go
```diff
@@ -0,0 +1,174 @@
...
+	Steps       []TestCaseSteps `json:"steps,omitempty"`
+}
+
+var (
+	qaseToken = os.Getenv(QaseTokenEnvVar)
+	Client    *qase.APIClient
+)
+
+// SetupQaseClient creates a new Qase client from the api token environment variable QASE_AUTOMATION_TOKEN
+func SetupQaseClient() {
```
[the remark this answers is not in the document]
rancher-max: Good call -- updated so that there is a mix of public and private functions now

rancher-max:
Made changes based on review. Changed the schema file to have "automated via" included in the description field so that I could confirm the update still worked. Confirmed that both create and update did still work. Output in terminal looks like:
```
INFO[0000] Uploading test case:
        Project: RANCHERINT
        Title: Test deploying fleet git repo on provisioned downstream cluster
        Description: Automated via TestGitRepoDeployment
        SuiteId: 5
        Automation: 2
        Steps: [{Create Rancher Instance   1} {Provision a kubernetes cluster  cluster name: testcluster1 2} {Create a new project and namespace  project name: fleet-test, namespace-name: fleet-testns 3} {Deploy a GitRepo object the gitRepo itself comes to an active state and the resources defined in the spec are created on the downstream cluster in the fleet-testns namespace /validation/fleet/schemas/gitrepo.yaml

apiVersion: fleet.cattle.io/v1alpha1
kind: GitRepo
metadata:
  name: test-gitrepo
  namespace: fleet-default
spec:
  repo: "https://github.com/rancher/fleet-examples"
  branch: "master"
  paths:
  - simple
  targetNamespace: "fleet-testns"
  targets:
  - clusterName: "testcluster1" 4}] 
INFO[0000] Uploading test case:
        Project: RANCHERINT
        Title: Example test case
        Description: This is an example description highlighting that you can include whatever you want in this raw text. It should be a fairly short description though and/or the associated automated test case.
        SuiteId: 19
        Automation: 2
        Steps: [{Example first action  Example data 1} {Example final action Example thing to validate here  2}]
```

validation/pipeline/qase/README.md
```diff
@@ -1,13 +1,19 @@
...
+  - [Schema Upload](#schema-upload)
+  - [Test Run](#test-run)
 
 ## Reporter
 Reporter retreives all test cases inorder to determine if said automation test exists or not. If it does not it will create the test case. There is a custom field for automation test name, so we can update results for existing tests. This is to determine if a pre-existing manual test case has been automated. This value should be the package and test name, ex: TestTokenTestSuite/TestPatchTokenTest1. It will then update the status of the test case, for a specifc test run provided. 
 
+## Schema Upload
+schemaupload searches for markdown files within a "schemas" folder on any package. These files should be named "[team_name]_schemas.md" and follow a **very specific structure**, where the top level heading corresponds to the Qase Project Code, the sub-heading from there is the Test Suite in Qase (of which there could be multiple sub-headings for different suites), and any sub-sub-headings are the test names. Any text that follows should be a short description of the test or the associated automated test, and then a markdown table for the steps. See /validation/fleet/schemas/... for an example.
```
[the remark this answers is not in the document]
rancher-max: I agree! I did it this way because we have a whole package for `schemas` so I was erroneously reading that if I just checked that it was in the path 😭 I am very open to suggestions for other names of the top level dir so that it's not redundant!

slickwarren (approved):
lgtm

slickwarren (approved):

PR #117, opened 2025-05-16
Istio appco automation

lscalabrini01:


validation/charts/appco/istio_test.go
```diff
@@ -0,0 +1,210 @@
...
+	i.T().Log("Installing SideCar Istio AppCo")
+	istioChart, logCmd, err := installIstioAppCo(client, i.cluster.ID, AppCoUsername, AppCoAccessToken, "")
+	require.NoError(i.T(), err)
+	require.True(i.T(), strings.Contains(logCmd, expectedDeployLog))
+	require.True(i.T(), istioChart.IsAlreadyInstalled)
+
+	i.T().Log("Uninstalling Istio AppCo")
+	istioChart, err = uninstallIstioAppCo(client, i.cluster.ID)
+	require.NoError(i.T(), err)
+	require.False(i.T(), istioChart.IsAlreadyInstalled)
```
floatingman, line 80: Can uninstall be put in a cleanup function?

lscalabrini01: Yes, I can

validation/charts/appco/README.md
```diff
@@ -0,0 +1,20 @@
...
+  "cleanup": false/optional,
+  "clusterName": "<cluster-to-run-test>"
+}
+```
+
+In your env vars, set the following:
+
+```bash
+set APPCO_USERNAME="<appco-username>"
+set APPCO_ACCESS_TOKEN="<appco-access-token>"
```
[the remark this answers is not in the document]
floatingman: Also, you might want to remove the set command, since it's Windows specific.

floatingman (dismissed):

validation/charts/appco/istio_test.go
lscalabrini01: I am facing closed sessions errors than I removed the uninstall from the cleanup function

validation/Jenkinsfile
```diff
@@ -117,6 +117,10 @@ node {
 
                 sh "./validation/configure.sh"
                 sh "./validation/build.sh"
+                if ("${ENV_VARIABLE}" != "null" && "${ENV_VARIABLE}" != "") {
```
slickwarren, line 120: can you give this a better name? there are many env vars, what does this one do? you also append the var to the end of the read in text, is that correct (line 122)?

validation/charts/appco/README.md
```diff
@@ -0,0 +1,32 @@
...
+  "clusterName": "<cluster-to-run-test>"
+}
+```
+
+In your env vars you can set the following:
+
+```bash
+APPCO_USERNAME="<appco-username>"
+APPCO_ACCESS_TOKEN="<appco-access-token>"
+```
```
slickwarren, line 20: is there a reason to use env vars vs. adding them as a struct somewhere and reading it in from the config? i.e. you would have something like 
```
rancher:
    host: caleb.qa.rancher.space
...
appco:
    username: caleb
    accessToken: minenotyours
```

validation/charts/appco/istio.go
```diff
@@ -0,0 +1,78 @@
...
+	extencharts "github.com/rancher/shepherd/extensions/charts"
+	"github.com/rancher/shepherd/extensions/kubectl"
+	namegen "github.com/rancher/shepherd/pkg/namegenerator"
+	"github.com/rancher/tests/actions/charts"
+	kubeapiNamespaces "github.com/rancher/tests/actions/kubeapi/namespaces"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	RancherIstioSecret string = "application-collection"
```
slickwarren, line 17: <nit> this is the `RancherIstioSecretName` not the secret itself

validation/charts/appco/istio.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+func createIstioNamespace(client *rancher.Client, clusterID string) error {
```
slickwarren, line 20: imo we shouldn't write helpers for a one liner, this would be pretty simple to just have inline.

validation/charts/appco/istio.go
```diff
@@ -0,0 +1,78 @@
...
+	_, err := kubeapiNamespaces.CreateNamespace(client, clusterID, namegen.AppendRandomString("testns"), charts.RancherIstioNamespace, "{}", map[string]string{}, map[string]string{})
+	return err
+}
+
+func createIstioSecret(client *rancher.Client, clusterID string, appCoUsername string, appCoToken string) (string, error) {
+	secretCommand := strings.Split(fmt.Sprintf("kubectl create secret docker-registry %s --docker-server=dp.apps.rancher.io --docker-username=%s --docker-password=%s -n %s", RancherIstioSecret, appCoUsername, appCoToken, charts.RancherIstioNamespace), " ")
+	logCmd, err := kubectl.Command(client, nil, clusterID, secretCommand, "")
```
slickwarren, line 27: nit; you could just return the command instead of saving to local vars first.

validation/charts/appco/istio.go
```diff
@@ -0,0 +1,78 @@
...
+	return logCmd, err
+}
+
+func installIstioAppCo(client *rancher.Client, clusterID string, appCoUsername string, appCoToken string, sets string) (*extencharts.ChartStatus, string, error) {
+	istioAppCoCommand := []string{
+		"sh", "-c",
+		fmt.Sprintf(`helm registry login dp.apps.rancher.io -u %s -p %s && helm install %s oci://dp.apps.rancher.io/charts/istio -n %s --set global.imagePullSecrets={%s} %s`, appCoUsername, appCoToken, charts.RancherIstioName, charts.RancherIstioNamespace, RancherIstioSecret, sets),
```
slickwarren, line 34: I think any/all of these command strings should be saved as a const for consistency.

validation/charts/appco/istio.go
slickwarren, line 1: is there something that checks / guarantees that kubectl and helm are already installed? this may be a redundant check, but might be worth adding.

validation/charts/appco/istio.go
```diff
@@ -0,0 +1,78 @@
...
+	return err
+}
+
+func createIstioSecret(client *rancher.Client, clusterID string, appCoUsername string, appCoToken string) (string, error) {
+	secretCommand := strings.Split(fmt.Sprintf("kubectl create secret docker-registry %s --docker-server=dp.apps.rancher.io --docker-username=%s --docker-password=%s -n %s", RancherIstioSecret, appCoUsername, appCoToken, charts.RancherIstioNamespace), " ")
+	logCmd, err := kubectl.Command(client, nil, clusterID, secretCommand, "")
+	return logCmd, err
+}
+
+func installIstioAppCo(client *rancher.Client, clusterID string, appCoUsername string, appCoToken string, sets string) (*extencharts.ChartStatus, string, error) {
```
slickwarren, line 31: nit; would be nice to include `wait` in the func name since it also waits on the chart to install to some capacity (line 43)

validation/charts/appco/istio_test.go
```diff
@@ -0,0 +1,220 @@
...
+	client  *rancher.Client
+	session *session.Session
+	cluster *clusters.ClusterMeta
+}
+
+func (i *IstioTestSuite) TearDownSuite() {
+	i.session.Cleanup()
+}
+
+func (i *IstioTestSuite) SetupSuite() {
```
slickwarren, line 33: you should make the secret creation and namespace creation part of setup vs. in every function. I wouldn't consider those as part of the named tests below, only as setup.

validation/charts/appco/istio_test.go
```diff
@@ -0,0 +1,220 @@
...
+	i.T().Logf("Creating %s namespace", charts.RancherIstioNamespace)
+	err = createIstioNamespace(client, i.cluster.ID)
+	require.NoError(i.T(), err)
+
+	i.T().Logf("Creating %s secret", RancherIstioSecret)
+	logCmd, err := createIstioSecret(client, i.cluster.ID, *AppCoUsername, *AppCoAccessToken)
+	require.NoError(i.T(), err)
+	require.True(i.T(), strings.Contains(logCmd, RancherIstioSecret))
+
+	ambientSets := `--set cni.enabled=true,ztunnel.enabled=true --set istiod.cni.enabled=false --set cni.profile=ambient,istiod.profile=ambient,ztunnel.profile=ambient`
```
slickwarren, line 90: save this in a const instead

validation/charts/appco/istio_test.go
```diff
@@ -0,0 +1,220 @@
...
+	i.T().Logf("Creating %s namespace", charts.RancherIstioNamespace)
+	err = createIstioNamespace(client, i.cluster.ID)
+	require.NoError(i.T(), err)
+
+	i.T().Logf("Creating %s secret", RancherIstioSecret)
+	logCmd, err := createIstioSecret(client, i.cluster.ID, *AppCoUsername, *AppCoAccessToken)
+	require.NoError(i.T(), err)
+	require.True(i.T(), strings.Contains(logCmd, RancherIstioSecret))
+
+	gatewaySets := `--set base.enabled=false,istiod.enabled=false --set gateway.enabled=true,gateway.namespaceOverride=default`
```
slickwarren, line 115: save this in a const instead

validation/charts/appco/istio_test.go
```diff
@@ -0,0 +1,220 @@
...
+	require.NoError(i.T(), err)
+	require.True(i.T(), strings.Contains(logCmd, RancherIstioSecret))
+
+	i.T().Log("Installing SideCar Istio AppCo")
+	istioChart, logCmd, err := installIstioAppCo(client, i.cluster.ID, *AppCoUsername, *AppCoAccessToken, "")
+	require.NoError(i.T(), err)
+	require.True(i.T(), strings.Contains(logCmd, expectedDeployLog))
+	require.True(i.T(), istioChart.IsAlreadyInstalled)
+
+	i.T().Log("Running In Place Istio AppCo Upgrade")
```
slickwarren, line 211: what makes this 'in place' vs. a normal upgrade?

validation/charts/appco/istio_test.go
```diff
@@ -0,0 +1,220 @@
...
+	gatewayNamespaceSets := `--set gateway.enabled=true,gateway.namespaceOverride=default`
+
+	i.T().Log("Installing Gateway Namespace Istio AppCo")
+	istioChart, logCmd, err := installIstioAppCo(client, i.cluster.ID, *AppCoUsername, *AppCoAccessToken, gatewayNamespaceSets)
+	require.NoError(i.T(), err)
+	require.True(i.T(), strings.Contains(logCmd, expectedDeployLog))
+	require.True(i.T(), istioChart.IsAlreadyInstalled)
+}
+
+func (i *IstioTestSuite) TestCanaryUpgrade() {
```
slickwarren, line 149: this appears to be `installIstioUpgradeWithCanary` or something similar, so that the name can reflect that
1. the chart will have a clean install as part of the test
2. upgrade will install / configure canary after the install is successful.

validation/charts/appco/README.md
lscalabrini01: We discussed this during one of our daily sessions, and using environment variables seemed like a better solution for passing sensitive information.

validation/charts/appco/istio_test.go
lscalabrini01: I can't add it to the suite because I need to delete the namespace to remove all Kubernetes objects created by the AppCo installation, naturally, I also need to recreate the namespace for each test. But I can add the namespace and the secret creation on the SetupTest

validation/charts/appco/istio.go
```diff
@@ -7,34 +7,37 @@ import (
...
-	kubeapiNamespaces "github.com/rancher/tests/actions/kubeapi/namespaces"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	RancherIstioSecret string = "application-collection"
+	ExpectedDeployLog                                  = "deployed"
+	IstioCanaryRevisionApp                             = "istiod-canary"
+	RancherIstioSecretName                      string = `application-collection`
+	IstioAmbientModeSet                         string = `--set cni.enabled=true,ztunnel.enabled=true --set istiod.cni.enabled=false --set cni.profile=ambient,istiod.profile=ambient,ztunnel.profile=ambient`
```
slickwarren, line 18: these don't need to be public afaik, since they are test-level they can just be private functions

validation/charts/appco/istio_test.go
```diff
@@ -56,18 +51,18 @@ func (i *IstioTestSuite) TestSideCarInstallation() {
...
-	err = createIstioNamespace(client, i.cluster.ID)
+	_, err = kubeapinamespaces.CreateNamespace(client, i.cluster.ID, namegen.AppendRandomString("testns"), charts.RancherIstioNamespace, "{}", map[string]string{}, map[string]string{})
 	require.NoError(i.T(), err)
 
-	i.T().Logf("Creating %s secret", RancherIstioSecret)
+	i.T().Logf("Creating %s secret", RancherIstioSecretName)
 	logCmd, err := createIstioSecret(client, i.cluster.ID, *AppCoUsername, *AppCoAccessToken)
 	require.NoError(i.T(), err)
-	require.True(i.T(), strings.Contains(logCmd, RancherIstioSecret))
+	require.True(i.T(), strings.Contains(logCmd, RancherIstioSecretName))
```
slickwarren, line 60: can't you move this to setup for all the tests you have?

validation/charts/appco/istio_test.go
slickwarren: hmm, @rancher-max thoughts on using in-place vs. leaving it out? putting it in makes it sound like a special case imo.

slickwarren (approved):

floatingman (approved):

validation/charts/appco/istio_test.go
rancher-max: It does make it sound like a special case, but I think it's fine to leave it in since that's how it is described in the documentation. Ref: https://docs.apps.rancher.io/reference-guides/istio/#upgrade-the-chart

PR #118, opened 2025-05-20
Updated the recurring runs Jenkinsfile to use the new qa-infra repo

igomez06:
Removed corral. Wrote some helper scripts, as well removed corral code that is no longer needed. Fixed two build error in the validation test side.

### DIFF
--- a/validation/pipeline/Jenkinsfile.recurring
+++ b/validation/pipeline/Jenkinsfile.recurring
@@ -2,6 +2,7 @@
 node {
     def rootPath = "/root/go/src/github.com/rancher/tests/"
     def workPath = "/root/go/src/github.com/rancher/tests/validation/"
+    def qaInfraWorkPath = "/root/go/src/github.com/rancher/qa-infra-automation"
     def jobName = "${JOB_NAME}"
     if (jobName.contains('/')) { 
       jobNames = jobName.split('/')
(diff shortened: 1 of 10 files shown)


validation/pipeline/Jenkinsfile.recurring
```diff
@@ -224,65 +185,36 @@ node {
...
-                        string(name: 'TEST_PACKAGE', value: "${env.TEST_PACKAGE}"),
-                        string(name: 'TAGS', value: "${env.TAGS}"),
-                        string(name: 'GOTEST_TESTCASE', value: "${env.GOTEST_TESTCASE} -p 1"),
-                      ]
-
-                      jobs["provisioning-test-validation"] = { build job: individualJob, parameters: goValidationParams }
-                    
-                    }
+                    jobs["test-provisioning"] = { build job: individualJob, parameters: goProvisioningParams }
+                    jobs["test-validation"] = { build job: individualJob, parameters: goValidationParams }
```
rancher-max, line 217: Probably should parameterize these job names in case they ever change for any reason.

validation/pipeline/scripts/build_qa_infra.sh
```diff
@@ -0,0 +1,94 @@
...
+export ANSIBLE_PRIVATE_KEY_FILE="$PRIVATE_KEY_FILE"
+
+# Apply the Terraform configuration
+terraform -chdir="$TERRAFORM_DIR" apply -auto-approve -var-file="$TFVARS_FILE"
+if [ $? -ne 0 ]; then
+    echo "Error: Terraform apply failed."
+    exit 1
+fi
+
+# --- RKE2 Playbook ---
```
rancher-max, line 40: I'm almost wondering if the playbooks should be its own separate script (and stage in the Jenkinsfile) so that it's easier to create other jobs in the future in case we want anything that just builds infra or just deploys on provided infra. Right now I don't think it's a big deal since it's currently built kind of coupled, but as we decouple to add more providers to the qa-infra-automation, it will make more sense to have these decouple as well.

validation/pipeline/scripts/build_qa_infra.sh
```diff
@@ -0,0 +1,94 @@
...
+    fi
+    echo "Terraform infrastructure destroyed successfully!"
+    exit 1
+fi
+
+# Export KUBECONFIG
+export KUBECONFIG="$KUBECONFIG_FILE"
+echo $KUBECONFIG
+
+# --- Rancher Playbook ---
```
rancher-max, line 79: Does this one need various attempts as well, or does it fail with multiple attempts anyway since it will likely deploy some resources into the cluster and need to destroy those on subsequent tries?

rancher-max (approved):
Left some various comments. There are also some pieces of commented out code that likely aren't necessary anymore at all and can just be removed. BUT nothing is blocking, so approving! Thank you for this!

validation/pipeline/Jenkinsfile.recurring
igomez06: Hey Max I will change this, when I refactor the job with the new shared library stuff that Ugur has been working on. I have task in the backlog for it. For this go around I just wanted to be functional.

validation/pipeline/scripts/build_qa_infra.sh
igomez06: Yes! Retries are built in.

caliskanugur (approved):

PR #122, opened 2025-06-03
Parallelism/Qase POC

susesgartner:
POC for parallelism on a local level and updates to the qase reporter to support reporting for local reporting.

Goals: 
1. Remove testify as it is no longer needed and doesn't provide much. 
2. Allow tests to be run in groups locally and in parallel via build tags and built in golang parallel functionality.
Note: Longterm this will make us agnostic to Jenkins/GH actions since we don't have to do any stringing together of tests in our runner of choice. 
3. Update the Qase reporter to match the changes in the previous steps.


rancher-max:
I think you did this as part of the other PR so this can likely be closed out now

PR #125, opened 2025-06-06
Add action for uploading schema files

rancher-max:
This is just a simple GH Action that runs the schemaupload on merge to main. It requires a change to be made to a schema to run at all.

See https://github.com/rancher-max/rancher-tests/actions/runs/15501356038/job/43649494777#step:4:12 for a working run. I have since updated the step names and removed the edit to the schema file that caused this to trigger.

### DIFF
--- a/.github/workflows/qase.yml
+++ b/.github/workflows/qase.yml
@@ -10,15 +10,15 @@ jobs:
   run-script:
     name: update-qase
     runs-on: ubuntu-latest
-    timeout-minutes: 1
+    timeout-minutes: 5
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: recursive
       
       - name: Setup go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
         with:
           go-version-file: "./go.mod"
           cache-dependency-path: "**/*.sum"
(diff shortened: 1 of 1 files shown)


.github/workflows/qase.yml
```diff
@@ -0,0 +1,29 @@
...
+      - '**/schemas/**'
+  
+jobs:
+  run-script:
+    name: update-qase
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
slickwarren, line 16: https://github.com/rancher/security-team/blob/main/docs/standards/rancher-gha-standards.md#security-recommendations

.github/workflows/qase.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches:
+      - 'main'
+    paths:
+      - '**/schemas/**'
+  
+jobs:
+  run-script:
+    name: update-qase
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
```
slickwarren, line 13: this seems pretty short, will this work for larger contributions too? or would 2 minutes be better?

.github/workflows/qase.yml
rancher-max: you're right... I was basing off of the FOSSA action though directly from rancher/rancher! 🤯 https://github.com/rancher/rancher/blob/main/.github/workflows/fossa.yml#L18

.github/workflows/qase.yml
rancher-max: good question. I don't know! Maybe I should set at 5 minutes. Shouldn't hurt anything 🤷

susesgartner (approved):

PR #126, opened 2025-06-10
Update creates.go

navsalehi:
This PR updates creates.go to fix flaky behavior in snapshot_s3_restore_test.go (issue [#1847](https://github.com/rancher/rancher/issues/1847)).
The test would sometimes restore a local snapshot on S3-enabled clusters, causing errors.
With this update, the test consistently selects and restores snapshots from S3 when S3 is enabled, ensuring reliable automation runs.


markusewalker (commented):
Overall, lgtm. Holding off on approving/merging until we get offline Jenkins jobs for RKE2/K3s S3 tests.

PR #131, opened 2025-06-12
Added test cases to Qase for AppCo Istio + Rancher + Fleet

lscalabrini01:
Resolves https://github.com/rancher/qa-tasks/issues/1861

### DIFF
--- a/validation/charts/appco/schemas/pit_schemas.md
+++ b/validation/charts/appco/schemas/pit_schemas.md
@@ -2,7 +2,7 @@
 
 ## Test Suite: AppCo
 
-### Tested AppCo Istio SideCar Mode installation on a provisioned downstream cluster
+### Install in SideCar Mode
 
 TestSideCarInstallation
 
@@ -18,7 +18,7 @@ TestSideCarInstallation
 
 ## Test Suite: AppCo
 
-### Tested AppCo Istio Ambient Mode installation on a provisioned downstream cluster
+### Install in Ambient Mode
 
 TestAmbientInstallation
 
@@ -34,7 +34,7 @@ TestAmbientInstallation
 
 ## Test Suite: AppCo
 
-### Tested AppCo Istio Standalone Mode installation on a provisioned downstream cluster
+### Install in Standalone Mode
 
 TestGatewayStandaloneInstallation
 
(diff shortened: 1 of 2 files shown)


floatingman (approved):
Looks good. Just one small knit. Is it possible to align the table cells? It's challenging to verify that the individual columns contain the correct data. This might be better handled automatically through a formatter or linter for Markdown, so I won't hold this PR up just for that.

floatingman:
> Looks good. Just one small knit. Is it possible to align the table cells? It's challenging to verify that the individual columns contain the correct data. This might be better handled automatically through a formatter or linter for Markdown, so I won't hold this PR up just for that.

Or we can just view the raw file, and GitHub will render it formatted for us.

rancher-max:
> Are these test cases automated yet? If not, then mark them as manual in Qase. Also, I wonder about the type of these tests, should they be marked as integration in Qase @rancher-max?

Caleb's PR has an update to allow marking them manual, but in the current state we don't have that option. I agree that we should do that, so for anything that isn't yet automated, change the description to "TBD" and it will get listed correctly after Caleb's PR merges. For type of test, we don't have anything yet for that.

validation/charts/appco/schemas/pit_schemas.md
```diff
@@ -0,0 +1,99 @@
...
+| ----------- | -------------------- | ------------ | ------------------------------ |
+| 1           | Config a downstream cluster running in rancher |   |   |
+| 2           | Create Namespace | Namespace name: istio-system |   |
+| 3           | Create Secret     | Secret name: application-collection |   |
+| 4           | Install Istio AppCo in SideCar mode | helm install <release-name> oci://dp.apps.rancher.io charts/istio -n istio-system --set global.imagePullSecrets={application-collection} |   |
+| 5           | Wait for Istio Deployments be running | kubectl get pods | All the pods should be running |
+
+---
+
+## Test Suite: AppCo
```
rancher-max, line 19: These additional `Test Suite` lines aren't required since it's all the same test suite.

rancher-max (changes requested):
One small change, otherwise everything else LGTM!

validation/charts/appco/schemas/pit_schemas.md
rancher-max: You can technically have it defined like this where the suite is explicit per test case, but it does also work with the suite defined only once and then all the sub test cases that belong within that suite!

lscalabrini01: done!

slickwarren (approved):

PR #136, opened 2025-06-23
update update-psa role template to use fixed name namespaces-psa

Priyashetty17:
Update the update-psa role template to use the fixed name 'namespaces-psa' based on recent changes. The updated RoleTemplate is as follows:
```yaml
apiVersion: management.cattle.io/v3
builtin: false
context: project
description: ''
displayName: Manage PSA Labels
external: false
hidden: false
kind: RoleTemplate
metadata:
  name: namespaces-psa
rules:
  - apiGroups:
      - management.cattle.io
    resources:
      - projects
    verbs:
      - updatepsa
```


brudnak (approved):
lgtm :shipit:

PR #132, opened 2025-06-12
adding harvester schemas

slickwarren:
also updated qase schemaUploader to allow for non-automated tests
if the description is TBD, it is not yet automated.

PR #140, opened 2025-06-24
Qase reporter refactor

susesgartner:
New Reporter:
- Adds the ability to report parameters for test cases.
- Adds support for using the reporter with tests running in parallel
- Adds support for mirroring qase tests between projects
- Updates current tests to utilize the schema uploader.
- Allows users to provide a lot more details of their test cases via qase upstream objects. (almost all qase test case settings are customizable)
- Fixes an issue with the old reporter that would cause multiple copies of the same test to be reported.
- Enables local Qase reporting

Updates:
- Updated the old schema uploader to utilize yaml (this makes it easier to utilize upstream objects)

As part of this PR the old Qase reporter should function normally and can be deprecated at a later date after all tests have a schema file.


rancher-max (commented):
I started reviewing and then got pulled into some other items so haven't fully reviewed yet, but wanted to get my comment out here at least for what I've seen so far. 

Is this requiring modifications in each test case as well? I see this chunk included often:
```
params, err := provisioning.GetProvisioningSchemaParams(tt.client, c.cattleConfigs[0])
require.NoError(c.T(), err)

err = qase.UpdateSchemaParameters(tt.name, params)
require.NoError(c.T(), err)
```
My understanding is that this is specifically for table tests. My concern is that it requires more understanding of a separate system as part of the test, and then what if that fails for some reason... does the test then fail? Can this be refactored to decouple the Qase piece from the test itself?

validation/provisioning/rke1/provisioning_node_driver_test.go
```diff
@@ -94,7 +94,7 @@ func (r *RKE1NodeDriverProvisioningTestSuite) TestProvisioningRKE1Cluster() {
 		client    *rancher.Client
 		runFlag   bool
 	}{
-		{"1 Node all roles " + provisioninginput.StandardClientName.String(), nodeRolesAll, r.standardUserClient, r.client.Flags.GetValue(environmentflag.Short) || r.client.Flags.GetValue(environmentflag.Long)},
+		{"RKE2_Node_Driver|etcd_cp_worker", nodeRolesAll, r.standardUserClient, r.client.Flags.GetValue(environmentflag.Short) || r.client.Flags.GetValue(environmentflag.Long)},
```
markusewalker, line 97: Was this a typo? This should be RKE1 for all of the test cases.

validation/provisioning/rke2/agent_customization_test.go
```diff
@@ -120,8 +121,8 @@ func (r *RKE2AgentCustomizationTestSuite) TestProvisioningRKE2ClusterAgentCustom
 		client       *rancher.Client
 		agent        string
 	}{
-		{"Custom Fleet Agent - Standard User", productionPool, r.standardUserClient, customAgents[0]},
-		{"Custom Cluster Agent - Standard User", productionPool, r.standardUserClient, customAgents[1]},
+		{"Custom_Fleet_Agent", productionPool, r.standardUserClient, customAgents[0]},
```
markusewalker, line 124: Should have RKE2 for both of the test cases as there is a K3s counterpart.

validation/snapshot/snapshot_additional_test.go
```diff
@@ -73,37 +70,31 @@ func (s *SnapshotAdditionalTestsTestSuite) TestSnapshotReplaceNodes() {
...
 	tests := []struct {
 		name         string
+		clusterType  string
 		etcdSnapshot *etcdsnapshot.Config
 		client       *rancher.Client
 	}{
-		{"Replace control plane nodes", controlPlaneSnapshotRestore, s.client},
-		{"Replace etcd nodes", etcdSnapshotRestore, s.client},
-		{"Replace worker nodes", workerSnapshotRestore, s.client},
+		{"RKE1_Replace_Control_Plane_Nodes", "rke1", controlPlaneSnapshotRestore, s.client},
```
markusewalker, line 77: Same comment/question earlier of needing to provide each cluster type.

validation/snapshot/snapshot_additional_test.go
```diff
@@ -118,34 +109,24 @@ func (s *SnapshotAdditionalTestsTestSuite) TestSnapshotRecurringRestores() {
 
 	tests := []struct {
 		name         string
+		clusterType  string
 		etcdSnapshot *etcdsnapshot.Config
 		client       *rancher.Client
 	}{
-		{"Restore snapshot 5 times", snapshotRestoreFiveTimes, s.client},
+		{"RKE1_Recurring_Restores", "rke1", snapshotRestoreFiveTimes, s.client},
```
markusewalker, line 116: Same comment/question earlier of needing to provide each cluster type.

validation/snapshot/snapshot_restore_k8s_upgrade_test.go
```diff
@@ -52,35 +49,25 @@ func (s *SnapshotRestoreK8sUpgradeTestSuite) TestSnapshotRestoreK8sUpgrade() {
 
 	tests := []struct {
 		name         string
+		clusterType  string
 		etcdSnapshot *etcdsnapshot.Config
 		client       *rancher.Client
 	}{
-		{"Restore Kubernetes version and etcd", snapshotRestoreK8sVersion, s.client},
+		{"RKE1_Restore_ETCD_K8sVersion", "-rancher", snapshotRestoreK8sVersion, s.client},
```
markusewalker, line 56: Same comment/question earlier of needing to provide each cluster type. Same for the rest of the snapshot tests.

markusewalker (commented):
Left a couple of comments. We can sync Monday to clear up some of my confusion. Other than that, awesome work!

validation/nodescaling/scale_replace_test.go
```diff
@@ -51,33 +52,36 @@ func (s *NodeReplacingTestSuite) TestReplacingNodes() {
...
+		nodeRoles    machinepools.NodeRoles
+		client       *rancher.Client
 	}{
-		{"control plane nodes", nodeRolesControlPlane, s.client},
-		{"etcd nodes", nodeRolesEtcd, s.client},
-		{"worker nodes", nodeRolesWorker, s.client},
+		{"RKE2_Node_Driver_Replace_Control_Plane", "rke2", nodeRolesControlPlane, s.client},
+		{"RKE2_Node_Driver_Replace_ETCD", "rke2", nodeRolesEtcd, s.client},
+		{"RKE2_Node_Driver_Replace_Worker", "rke2", nodeRolesWorker, s.client},
+		{"K3S_Node_Driver_Replace_Control_Plane", "k3s", nodeRolesControlPlane, s.client},
```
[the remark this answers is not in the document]
susesgartner: Yes, you would need to provide each cluster type. i.e only rke2 table tests will run with the rke2 cluster and only k3s table tests will run with k3s

validation/provisioning/rke1/provisioning_node_driver_test.go
susesgartner: Yep. Good catch

validation/provisioning/rke2/agent_customization_test.go
susesgartner: There currently is not a k3s counterpart in the k3s directory. Unless i'm missing it 😓

markusewalker: You aren't, my bad, I thought it existed 😬

markusewalker (dismissed):

Josh-Diamond (dismissed):

Josh-Diamond (dismissed):

markusewalker (dismissed):

markusewalker (dismissed):

actions/qase/qase.go
```diff
@@ -0,0 +1,195 @@
...
+		if err == nil {
+			logrus.Info("Updating test case:\n\tProject: ", project, "\n\tTitle: ", tc.Title, "\n\tSuiteId: ", tc.SuiteId)
+			var qaseTest upstream.TestCaseUpdate
+			qaseTest.Title = tc.Title
+			qaseTest.SuiteId = tc.SuiteId
+			qaseTest.Description = tc.Description
+			qaseTest.Type_ = tc.Type_
+			qaseTest.Priority = tc.Priority
+			qaseTest.IsFlaky = tc.IsFlaky
+			qaseTest.Automation = int32(2)
```
rancher-max, line 139: There was a nice update to make this not always say automated -- can we add this?

ref: https://github.com/rancher/tests/pull/132/files#diff-1b2c3b1e83b3ebea187affbb7a36cfc882e7866c791de1692ea199f017de4811

actions/qase/schema.go
```diff
@@ -0,0 +1,146 @@
...
+	err := filepath.Walk(basePath, func(path string, info os.FileInfo, err error) error {
+		if strings.Contains(info.Name(), schemas) {
+			var projectSchema TestProjectSchema
+
+			fileContent, err := os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+
+			err = yaml.Unmarshal(fileContent, &projectSchema)
```
rancher-max, line 27: This err may be unused/dangling. There are probably the original schemas file we added when this was supported with .md format instead of .yaml that is causing this err to actually exist, I would guess

actions/qase/schema.go
```diff
@@ -0,0 +1,146 @@
...
+
+// UpdateSchemaParameters updates the parameters of a test's schema file
+func UpdateSchemaParameters(testName string, params []upstream.Params) error {
+	_, schemaFilepath, _, _ := runtime.Caller(1)
+	packagePath := filepath.Dir(schemaFilepath)
+
+	qaseProjectSchemas, err := GetSchemas(packagePath)
+	if len(qaseProjectSchemas) > 1 {
+		return fmt.Errorf("%s contains multiple schema files", packagePath)
+	}
```
rancher-max, line 69: couldn't a specific project have multiple schemas? For example, the provisioning package may have test cases required by Harvester, Hostbusters, and PIT, which would all need separate schemas files for the respective Qase projects.

actions/qase/testresult/testresult.go
```diff
@@ -10,9 +10,10 @@ type GoTestOutput struct {
 	Elapsed string `json:"Elapsed" yaml:"Elapsed"`
 }
 
-// GoTestCase is the struct used for sending the appropriate API call to Qase
-type GoTestCase struct {
+// GoTestResult is the struct for holding test results
+type GoTestResult struct {
```
rancher-max, line 14: nit: there are a few local variables that are using this renamed type that are still `testcase`

susesgartner: Your right. Hopefully I got them all with the latest commit

actions/qase/schema.go
susesgartner: Fixed

actions/qase/schema.go
rancher-max: Does this account for different Qase projects though? We have some teams who aren't using the RM project

susesgartner: Yes it does, I tested it using a non-RM project

rancher-max: that's reporter though not the uploader

susesgartner: Ah, I misunderstood. But you can use any project for either the uploader OR the reporter.

susesgartner: Currently if you need a test in multiple projects I don't have any code in place for that but i'm pretty sure that was the case with the previous uploader as well so i'm not really changing the status quo here.

susesgartner: After considering this a bit more. I don't really have a mechanism for uploading an individual test to several projects at once. I could potentially change the project field to specify multiple projects, the only downside to doing this is that it would be handled at a schema files level which would mean you couldn't really pick and choose 1 or 2 tests from a schema file, it would be all or nothing. Previously you could get around this with the old schema uploader by duplicating the tests into multiple different schema files (not ideal either). Is this something you view as super important or can I handle this in a follow up PR? @rancher-max

rancher-max: In the current state, and with the GH action, it will go and try to update so yeah I would prefer if this behavior didn't change 😬

markusewalker (dismissed):

rancher-max (dismissed):
Alright I think this is fine! Thank you for the changes!

markusewalker (approved):

rancher-max (approved):

PR #145, opened 2025-07-01
Review changes

lscalabrini01:


rancher-max (approved):
Interesting, so adding the diff stuff into the fleet yaml makes it so that the pilot image isn't necessary anymore? This is really clean on this side! Can you also update the schema if applicable to make sure it points to the correct location once that fleet-examples PR is merged?

floatingman (approved):
LGTM

PR #147, opened 2025-07-03
Add 'custodian' support for linode

slickwarren:
Introducing some bash scripts to act as a custodian for linode. Deletes any node older than 2 days, that doesn't have the DONOTDELETE_KEYS

Added any new keys to our store already. For user management, only uses existing keys so no change(s) to process


scripts/custodian/Dockerfile
```diff
@@ -28,7 +28,9 @@ ENV DONOTDELETE_KEYS=$DONOTDELETE_KEYS
 ENV CUSTODIAN_YAML=aws.yaml
 
 ENV GOOGLE_APPLICATION_CREDENTIALS=google_credentials.json
-ENV GOOGLE_CLOUD_PROJECT=ei-container-platform-qa
+
+ARG GOOGLE_CLOUD_PROJECT
+ENV GOOGLE_CLOUD_PROJECT=GOOGLE_CLOUD_PROJECT
```
rancher-max, line 33: I don't know why this looks weird to me... is the ENV supposed to be `GOOGLE_CLOUD_PROJECT=${GOOGLE_CLOUD_PROJECT}`?

scripts/custodian/Jenkinsfile
```diff
@@ -84,7 +87,11 @@ node("vsphere-vpn-1") {
               }
             }
             stage('Run Docker Image for AWS,Azure, and GCP') {
```
rancher-max, line 89: nit: change stage name to account for Linode as well

rancher-max (approved):
One comment (more of a question) but otherwise LGTM

scripts/custodian/Dockerfile
slickwarren: ah, was missing the $

scripts/custodian/Jenkinsfile
slickwarren: ack

rancher-max (dismissed):

floatingman (changes requested):

floatingman (approved):

PR #148, opened 2025-07-08
Fix VAI enabled detection when spec.value is nil

brudnak:
### Problem
VAI is now enabled by default in Rancher, but our `isVaiEnabled` function was treating `spec.value: nil` as disabled. This caused tests to incorrectly detect VAI as disabled on fresh clusters.

### Solution
Updated `isVaiEnabled` to check `status.default` when `spec.value` is nil or missing.

### Changes
- When `spec.value` is nil → use `status.default` (instead of returning false)
- Added structured logging to see what's happening
- Handle edge cases where status might be missing

### Examples

**Before:** Fresh cluster with VAI default enabled
spec.value: null
status.default: true
Old function returns: false ❌

**After:** Same cluster
Feature: ui-sql-cache
  spec.value: <nil>
  status.default: true
  VAI enabled: true (spec.value is nil, using default) ✅

### Why it matters
- Tests were trying to enable VAI when it was already enabled
- Cleanup was reverting to the wrong state
- This aligns with how Rancher actually interprets feature flags

### DIFF
--- a/validation/steve/vai/vai.go
+++ b/validation/steve/vai/vai.go
@@ -28,30 +28,56 @@ type SupportedWithVai interface {
 
 func isVaiEnabled(client *rancher.Client) (bool, error) {
 	managementClient := client.Steve.SteveType("management.cattle.io.feature")
-	steveCacheFlagResp, err := managementClient.ByID(uiSQLCacheResource)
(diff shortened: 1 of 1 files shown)


fillipehmeireles (approved):
lgtm!

PR #152, opened 2025-07-09
fix limit continue tests

brudnak:
limit continue not supported when vai is enabled
fixed a bug in the test so it's working properly
now when vai is disabled


fillipehmeireles (approved):
lgtm!

PR #154, opened 2025-07-14
Deprecate RKE1 test for 2.12

Priyashetty17:
Since RKE1 clusters are no longer supported in v2.12.0, deprecate tests specific to RKE1.


brudnak (approved):
lgtm

PR #158, opened 2025-07-17
Exploring test cases for Elemental and Rancher Manager

lscalabrini01:


validation/elemental/schemas/elemental_schemas.yaml
```diff
@@ -0,0 +1,83 @@
...
+          expectedresult: "NeuVector chart reaches Active state"
+
+    - title: "Test provisioning Elemental cluster on Rancher environment with Harvester"
+      description: "Validate that an Elemental cluster can be provisioned on a Rancher environment using Harvester as the infrastructure"
+      automation: 2
+      steps:
+        - position: 1
+          action: "Create Proxy Rancher"
+          data: ""
+          expectedresult: "Proxied Rancher instance is up and accessible"
```
slickwarren, line 79: why does this have to be proxied?

slickwarren (approved):
are there any elemental-only tests we should add? 

I would think all of these have automation coverage, if the test is given the elemental cluster.  So I'm wondering how we will write these tests?

slickwarren (approved):

PR #162, opened 2025-07-22
update harvester job to use qa-infra repo

slickwarren:
* added helper scripts:
  * dockerfile.infra -> installs ansible, gotestsum, and tofu with support (symlink) for terraform commands
    * support for building the dev version if 0.0.0-dev is specified, harvester/rancher2 providers only
* update e2e harvester job to use qa-infra-automation instead of tfp-automation
  * converts yaml files from jenkins environment -> tfvars (necessary to make variable overrides easier)
  * replaces harvester VM provisioning + rancher HA install
  * introduces downstream cluster provisioning via infra framework in order to run tests later on a single, specified cluster


markusewalker (approved):

scripts/tofu/dev-provider.sh
```diff
@@ -0,0 +1,48 @@
...
+  PLATFORM="darwin_arm64"
+else
+  PLATFORM="linux_amd64"
+fi
+
+# Download binary
+DIR=~/.terraform.d/plugins/terraform.local/${PROVIDER}/${PROVIDER}/${VERSION}/${PLATFORM}
+mkdir -p $DIR
+(umask u=rwx,g=rwx,o=rwx && mkdir -p $DIR)
+git clone $GITURL provider-clone
```
floatingman, line 30: Not a huge deal, but it would be a good idea to remove this cloned directory at the end of this script.
Take a look at the `trap` function in bash.
You can do something like `trap 'rm -rf provider-clone' EXIT` after the `set -e` to remove the cloned directory after this script runs.

validation/Dockerfile.e2e
```diff
@@ -51,19 +57,27 @@ RUN TOFU_ZIP="tofu_${TOFU_VERSION}_linux_amd64.zip" && \
...
-RUN wget -qO ./yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/local/bin/yq && \
-    chmod +x /usr/local/bin/yq
+
+RUN if [[ "$RANCHER2_PROVIDER_VERSION" == *"-dev"* ]]; then \
+      chmod +x ./scripts/tofu/dev-provider.sh && ./scripts/tofu/dev-provider.sh rancher2 "https://github.com/rancher/terraform-provider-rancher2.git" ; \
+    fi; \
+    if [[ "$HARVESTER_PROVIDER_VERSION" == *"-dev"* ]]; then \
+      chmod +x ./scripts/tofu/dev-provider.sh && ./scripts/tofu/dev-provider.sh harvester "https://github.com/harvester/terraform-provider-harvester.git" ; \
+    fi; \
+    wait
```
floatingman, line 80: I think the `SHELL ["/bin/bash", "-c"] needs to be moved before this block

floatingman (changes requested):
Looks good overall. Just a few suggestions.

floatingman (dismissed):

hamistao (approved):
Some comments and questions, but looks good overall.

validation/pipeline/scripts/build_qa_infra.sh
```diff
@@ -1,30 +1,103 @@
 #!/bin/bash
 
-cd  /root/go/src/github.com/rancher/qa-infra-automation
+show_help() {
```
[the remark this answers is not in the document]
slickwarren: it gets used if you use the -h flag when calling the script

PR #166, opened 2025-07-24
Fixing node annotation tests to match UI

brudnak:
Updating node annotation tests to match API calls that are made when doing these same actions via the UI.


fillipehmeireles (approved):
Lgtm!

PR #169, opened 2025-07-29
RKE2 package re-work for recurring runs.

susesgartner:
Summary: The goal of this PR is to prepare our automation to be added to recurring runs by shortening the configs, allowing parallelism across and remove various components that were no longer necessary.

Changes:
- Removed Testify from all tests in the rke2 package (testify assertions is an exception)
- Changed the build tags to what we are going to be using for our recurring runs
- Added parallelism at the test and table test levels so we can run multiple tests and their table tests at the same time from a single command
- Added a defaults system that allows for non-sensitive fields to be ignored if the user chooses.
- Renamed the tests to remove redundancy.
- Split the dynamic tests out from the release tests (this was needed due to build tag constraints)
- Removed the hard coded load of the EC2Configs from the custom cluster and updated all affected tests (this was done because it was bypassing defaults)


results.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
markusewalker, line 1: Make sure that this is deleted, probably here by accident.

validation/provisioning/rke2/custom_test.go
```diff
@@ -0,0 +1,139 @@
+//go:build recurring
```
markusewalker, line 1: For the build tag, let's make sure we have the choice of `validation` as well. So should be `//go:build validation || recurring`.

validation/provisioning/rke2/dynamic_node_driver_test.go
```diff
@@ -0,0 +1,129 @@
+//go:build dynamic
```
markusewalker, line 1: Leaving the comment here, but not specific to this file. With the edits that you're making to these files, can you make sure that any changes needed to the respective READMEs are made?

validation/provisioning/rke2/custom_test.go
```diff
@@ -0,0 +1,139 @@
...
+		{"RKE2_Custom|etcd|cp|worker|windows", r.standardUserClient, nodeRolesDedicatedWindows, true},
+		{"RKE2_Custom|3_etcd|2_cp|3_worker", r.standardUserClient, nodeRolesStandard, false},
+	}
+	for _, tt := range tests {
+		clusterConfig := new(clusters.ClusterConfig)
+		operations.LoadObjectFromMap(defaults.ClusterConfigKey, r.cattleConfig, clusterConfig)
+
+		clusterConfig.MachinePools = tt.machinePools
+
+		t.Cleanup(func() {
```
markusewalker, line 108: Is this intended to be here so that at the start of each test, the prior cluster is cleaned up? If so, won't this run the first test case when there is no cluster initially provisioned?

validation/provisioning/rke2/defaults/defaults.yaml
```diff
@@ -0,0 +1,81 @@
+rancher:
```
markusewalker, line 1: If not already added, might be good to add a section in the READMEs explaining what these are.

validation/provisioning/rke2/hardened_test.go
```diff
@@ -1,11 +1,12 @@
-//go:build (validation || sanity) && !infra.any && !infra.aks && !infra.eks && !infra.rke2k3s && !infra.gke && !infra.rke1 && !cluster.any && !cluster.custom && !cluster.nodedriver && !extended && !stress
+//go:build recurring
```
markusewalker, line 1: This comment can be taken for all of the subsequent tests. Can you update to: `//go:build validation || recurring`?

markusewalker (commented):
Left some comments, but overall the functionality is good.

Josh-Diamond (commented):
Initial review LGTM -- will await requested changes from other reviewers before approving

validation/provisioning/rke2/custom_test.go
susesgartner: t.Cleanup doesn't run until the test is finished. So while the t.Cleanup function is defined here it won't run until after the thread finishes.

slickwarren (approved):
confirming this is OK from PIT crew perspective 👍🏼

markusewalker (approved):
LGTM.

PR #172, opened 2025-07-30
Update stable branch with current main

rancher-max:


slickwarren:
based on some previous discussions, can we instead start cutting versions instead of having a stable branch?

rancher-max:
ah yeah that's a good idea actually. I will close this out :)

PR #176, opened 2025-07-31
K3s package rework

susesgartner:
PR to rework the k3s package to use similar logic to the rke2 package.

Changes: 
1. All the changes listed in https://github.com/rancher/tests/pull/169 .
2. Updated logic in both RKE2/K3s packages to utilize the standard client helper function.


Josh-Diamond (approved):

markusewalker (approved):

validation/provisioning/k3s/README.md
```diff
@@ -69,170 +204,215 @@ clusterConfig:
       folder: ""
       region: "us-east-2"
       skipSSLVerify: true
+  
+  clusterAgent:                        # change this to fleetAgent for fleet agent
+  appendTolerations:
```
slickwarren, line 209: looks like there's an indent missing here

validation/provisioning/k3s/defaults/defaults.yaml
```diff
@@ -0,0 +1,80 @@
...
+      volumeSize: 100
+      roles: ["etcd", "controlplane", "worker"]
+
+#Required for tests that utilize custom clusters
+sshPath:
+   sshPath: "<>"
+
+awsCredentials:
+  secretKey: "<>"
+  accessKey: "<>"
```
slickwarren, line 68: does it make sense to have these in 'defaults' if they don't have a default value set?

validation/provisioning/k3s/hardened_test.go
```diff
@@ -0,0 +1,138 @@
...
+	nodeRolesStandard[1].MachinePoolConfig.Quantity = 2
+	nodeRolesStandard[2].MachinePoolConfig.Quantity = 3
+
+	tests := []struct {
+		name            string
+		client          *rancher.Client
+		machinePools    []provisioninginput.MachinePools
+		scanProfileName string
+	}{
+		{"K3S_CIS_1.9_Profile|3_etcd|2_cp|3_worker", k.standardUserClient, nodeRolesStandard, "k3s-cis-1.9-profile"},
```
slickwarren, line 80: for CIS profile, I doesn't that change somewhat often? If so, wouldn't it make sense to make this less hardcoded in some way, i.e. a regex on the profile version or something?

validation/provisioning/k3s/node_driver_test.go
```diff
@@ -0,0 +1,110 @@
...
+	client, err := rancher.NewClient("", testSession)
+	assert.NoError(t, err)
+	k.client = client
+
+	k.cattleConfig = config.LoadConfigFromFile(os.Getenv(config.ConfigEnvironmentKey))
+
+	k.cattleConfig, err = defaults.LoadPackageDefaults(k.cattleConfig, "")
+	assert.NoError(t, err)
+
+	k.cattleConfig, err = defaults.SetK8sDefault(k.client, "k3s", k.cattleConfig)
```
slickwarren, line 46: do we have consts for these strings?

validation/provisioning/k3s/node_driver_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	tests := []struct {
+		name         string
+		machinePools []provisioninginput.MachinePools
+		client       *rancher.Client
+	}{
+		{"K3S_Node_Driver|etcd_cp_worker", nodeRolesAll, k.standardUserClient},
+		{"K3S_Node_Driver|etcd_cp|worker", nodeRolesShared, k.standardUserClient},
+		{"K3S_Node_Driver|etcd|cp|worker", nodeRolesDedicated, k.standardUserClient},
+		{"K3S_Node_Driver|3_etcd|2_cp|3_worker", nodeRolesStandard, k.standardUserClient},
```
slickwarren, line 76: do we have consts for these strings? or in this case does it make sense not to use a const?

validation/provisioning/k3s/node_driver_test.go
```diff
@@ -0,0 +1,110 @@
+//go:build validation || recurring
```
slickwarren, line 1: how are you guys using tags, 'recurring' seems too vague imo. Do you have some cadence that you run the recurring tests on? If so, do you run the same set (everything with 'recurring' tag) every time, or a subset of these?

validation/provisioning/k3s/psact_test.go
```diff
@@ -7,73 +7,56 @@ import (
...
+	assert.NoError(t, err)
 
-	k.standardUserClient = standardUserClient
+	return k
 }
 
-func (k *K3SPSACTTestSuite) TestK3SPSACTNodeDriverCluster() {
+func TestPSACT(t *testing.T) {
+	t.Parallel()
+	k := psactSetup(t)
```
slickwarren, line 58: so you're still using testing.T, but not using the builtin setupSuite. Why?

slickwarren (dismissed):
none of my comments are blocking, but I do have some clarifying questions for my own understanding and some comments on syntax/formatting. LGTM for pit team

validation/provisioning/k3s/node_driver_test.go
susesgartner: In this case it wouldn't make sense since these are the names of test cases so it kind of makes sense to keep them in the table test (especially because they aren't reused anywhere else)

validation/provisioning/k3s/node_driver_test.go
susesgartner: For now we are going to run all of them and we are planning on a weekly/bi-weekly cadence. As we continue to upgrade our recurring runs we might change the tags later but that should be easy to change down the road.

validation/provisioning/k3s/defaults/defaults.yaml
susesgartner: I just added them to indicate that those fields are required for the objects in question. I suppose maybe I could replace <> with <required> if that would make it clearer.

PR #263, opened 2025-09-19
Add RKE2 IPv6 custom cluster test

markusewalker:
### Description
As part of broader IPv6 testing, this is an initial PR to introduce provisioning RKE2 IPv6 custom clusters. While the automation can handle node driver and some other existing tests, they all have known bugs against them. So for now, just creating the custom cluster test for RKE2 as that is passing.

Once some of the rancher bugs have been fixed, more tests will be added to further beef up IPv6 testing which will ultimately be put in workflows as discussed with the team offline.


slickwarren (commented):
gomod changes are approved

PR #345, opened 2025-11-07
use dynamic k8s version

Priyashetty17:
Automatically determine the correct k8s version for each Rancher instance, reducing manual inputs to the workflows.


joesims22 (approved):
LGTM

PR #357, opened 2025-11-12
add p0 support back for existing cluster tests

slickwarren:
This adds a new file to test various areas on existing clusters, using the rancher.clusterName field from configs.

### DIFF
--- a/validation/certificates/rke2k3s/cert_rotation_existing_cluster_test.go
+++ b/validation/certificates/rke2k3s/cert_rotation_existing_cluster_test.go
@@ -14,7 +14,6 @@ import (
 	"github.com/rancher/shepherd/pkg/session"
 	"github.com/rancher/tests/actions/logging"
 	"github.com/rancher/tests/actions/provisioning"
-	"github.com/rancher/tests/actions/qase"
 	"github.com/rancher/tests/actions/workloads/pods"
 	"github.com/rancher/tests/validation/certificates"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


validation/certificates/rke2k3s/cert_rotation_existing_cluster_test.go
```diff
@@ -0,0 +1,92 @@
+//go:build (validation || infra.rke2k3s || recurring || cluster.any || stress) && !infra.any && !infra.aks && !infra.eks && !infra.gke && !infra.rke1 && !sanity && !extended
```
susesgartner: This has the recurring tag in it which will cause problems with our recurring runs.

validation/nodescaling/rke2k3s/auto_replace_existing_cluster_test.go
```diff
@@ -0,0 +1,92 @@
+//go:build (validation || extended) && !infra.any && !infra.aks && !infra.eks && !infra.gke && !infra.rke2k3s && !cluster.any && !cluster.custom && !cluster.nodedriver && !sanity && !stress
```
susesgartner: Same as above

validation/certificates/rke2k3s/cert_rotation_existing_cluster_test.go
```diff
@@ -0,0 +1,92 @@
...
+	require.NoError(c.T(), err)
+
+}
+
+func (c *CertRotationExistingClusterTestSuite) TestCertRotationExistingCluster() {
+	tests := []struct {
+		name      string
+		clusterID string
+	}{
+		{"ExistingCluster_Certificate_Rotation", c.clusterObject.ID},
```
susesgartner: Are you intending to use this with V1 or V2 reporting?

slickwarren: I was just using whatever was supported by the previous test

PR #361, opened 2025-11-13
Reset settings to default values in UserRetention tests

joesims22:
This PR modifies the teardown suite in the UserRetention tests to reset the settings that are modified throughout the tests to their default values.


validation/auth/userretention/userretention.go
```diff
@@ -54,6 +55,35 @@ func setupUserRetentionSettings(client *rancher.Client, disableAfterValue string
 	return nil
 }
 
+func resetUserRetentionSettings(client *rancher.Client) error {
```
Priyashetty17: authUserSessionTTLMinutes is missing, can you add that as well?

PR #381, opened 2025-11-19
Improve Qase reports

lscalabrini01:


validation/pipeline/qase/reporter-v2/main.go
```diff
@@ -57,10 +60,15 @@ func main() {
...
 
-		err = reportTestQases(client, int32(runID))
+		err = reportTestQases(qaseService, int32(runID))
 		if err != nil {
 			logrus.Error("error reporting: ", err)
 		}
+
+		err = completeQaseReport(qaseService, projectIDEnvVar, int32(runID))
+		if err != nil {
+			logrus.Error("error update reporting: ", err)
```
slickwarren: why are these errors only logged, not fatals?

lscalabrini01: I guess that’s because they don’t want the job to fail

floatingman (approved):

slickwarren (approved):

PR #389, opened 2025-11-21
Add first Longhorn tests

hamistao:
https://github.com/rancher/qa-tasks/issues/2006

This adds the first automated Longhorn interoperability tests, namely:

- Longhorn Installation with Custom Configuration
- Fresh Longhorn Installation via Rancher App Catalog
- Create Longhorn volume through Rancher Workloads
- Test RBAC integration with Longhorn
- Test scaling a StatefulSet with a Longhorn PVC template

### DIFF
--- a/validation/longhorn/longhorn_test.go
+++ b/validation/longhorn/longhorn_test.go
@@ -1,0 +1,205 @@
+//go:build validation || pit.daily
+
+package longhorn
+
+import (
+	"slices"
+	"testing"
+
+	"github.com/rancher/shepherd/clients/rancher"
+	"github.com/rancher/shepherd/clients/rancher/catalog"
+	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	steveV1 "github.com/rancher/shepherd/clients/rancher/v1"
+	shepherdCharts "github.com/rancher/shepherd/extensions/charts"
+	"github.com/rancher/shepherd/extensions/clusters"
+	shepherdPods "github.com/rancher/shepherd/extensions/workloads/pods"
+	"github.com/rancher/shepherd/pkg/namegenerator"
+	"github.com/rancher/shepherd/pkg/session"
+	"github.com/rancher/tests/actions/charts"
+	"github.com/rancher/tests/actions/kubeapi/volumes/persistentvolumeclaims"
+	namespaceActions "github.com/rancher/tests/actions/namespaces"
+	"github.com/rancher/tests/actions/projects"
+	"github.com/rancher/tests/actions/rbac"
+	"github.com/rancher/tests/actions/storage"
+	"github.com/rancher/tests/actions/workloads/pods"
+	"github.com/rancher/tests/actions/workloads/statefulset"
+	"github.com/rancher/tests/interoperability/longhorn"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type LonghornTestSuite struct {
+	suite.Suite
+	client             *rancher.Client
+	session            *session.Session
+	longhornTestConfig longhorn.TestConfig
+	cluster            *clusters.ClusterMeta
+	project            *management.Project
+	payloadOpts        charts.PayloadOpts
+}
+
+func (l *LonghornTestSuite) TearDownSuite() {
+	l.session.Cleanup()
+}
+
+func (l *LonghornTestSuite) SetupSuite() {
+	l.session = session.NewSession()
+
+	client, err := rancher.NewClient("", l.session)
+	require.NoError(l.T(), err)
+	l.client = client
+
+	l.cluster, err = clusters.NewClusterMeta(client, client.RancherConfig.ClusterName)
+	require.NoError(l.T(), err)
+
+	l.longhornTestConfig = *longhorn.GetLonghornTestConfig()
+
+	projectConfig := &management.Project{
+		ClusterID: l.cluster.ID,
+		Name:      l.longhornTestConfig.LonghornTestProject,
+	}
+
+	l.project, err = client.Management.Project.Create(projectConfig)
+	require.NoError(l.T(), err)
+
+	chart, err := shepherdCharts.GetChartStatus(l.client, l.cluster.ID, charts.LonghornNamespace, charts.LonghornChartName)
+	require.NoError(l.T(), err)
+
+	if !chart.IsAlreadyInstalled {
+		// Get latest versions of longhorn
+		latestLonghornVersion, err := l.client.Catalog.GetLatestChartVersion(charts.LonghornChartName, catalog.RancherChartRepo)
+		require.NoError(l.T(), err)
+
+		payloadOpts := charts.PayloadOpts{
+			Namespace: charts.LonghornNamespace,
+			Host:      l.client.RancherConfig.Host,
+			InstallOptions: charts.InstallOptions{
+				Cluster:   l.cluster,
+				Version:   latestLonghornVersion,
+				ProjectID: l.project.ID,
+			},
+		}
+
+		l.T().Logf("Installing Lonhgorn chart in cluster [%v] with latest version [%v] in project [%v] and namespace [%v]", l.cluster.Name, l.payloadOpts.Version, l.project.Name, l.payloadOpts.Namespace)
+		err = charts.InstallLonghornChart(l.client, payloadOpts, nil)
+		require.NoError(l.T(), err)
+	}
+}
+
+func (l *LonghornTestSuite) TestRBACIntegration() {
+	cluster, err := l.client.Management.Cluster.ByID(l.cluster.ID)
+	require.NoError(l.T(), err)
+
+	project, namespace, err := projects.CreateProjectAndNamespaceUsingWrangler(l.client, l.cluster.ID)
+	require.NoError(l.T(), err)
+	l.T().Logf("Created project: %v", project.Name)
+
+	projectUser, projectUserClient, err := rbac.AddUserWithRoleToCluster(l.client, rbac.StandardUser.String(), rbac.ProjectMember.String(), cluster, project)
+	require.NoError(l.T(), err)
+	l.T().Logf("Created user: %v", projectUser.Username)
+
+	readOnlyUser, readOnlyUserClient, err := rbac.AddUserWithRoleToCluster(l.client, rbac.StandardUser.String(), rbac.ReadOnly.String(), cluster, project)
+	require.NoError(l.T(), err)
+	l.T().Logf("Created user: %v", readOnlyUser.Username)
+
+	storageClass, err := storage.GetStorageClass(l.client, l.cluster.ID, l.longhornTestConfig.LonghornTestStorageClass)
+	require.NoError(l.T(), err)
+
+	l.T().Log("Create and delete volume with admin user")
+	require.NoError(l.T(), storage.CreateAndDeleteVolume(l.client, l.cluster.ID, namespace.Name, storageClass))
+
+	l.T().Log("Create and delete volume with project user")
+	require.NoError(l.T(), storage.CreateAndDeleteVolume(projectUserClient, l.cluster.ID, namespace.Name, storageClass))
+
+	l.T().Log("Attempt to create and delete volume with project user on the wrong project")
+	require.Error(l.T(), storage.CreateAndDeleteVolume(projectUserClient, l.cluster.ID, charts.LonghornNamespace, storageClass))
+
+	l.T().Log("Attempt to create and delete volume with read-only user")
+	require.Error(l.T(), storage.CreateAndDeleteVolume(readOnlyUserClient, l.cluster.ID, namespace.Name, storageClass))
+}
+
+func (l *LonghornTestSuite) TestScaleStatefulSetWithPVC() {
+	namespaceName := namegenerator.AppendRandomString("lhsts")
+	namespace, err := namespaceActions.CreateNamespace(l.client, namespaceName, "{}", map[string]string{}, map[string]string{}, l.project)
+	require.NoError(l.T(), err)
+	l.T().Logf("Created namespace %s", namespaceName)
+
+	podTemplate := pods.CreateContainerAndPodTemplate()
+	statefulSet, err := statefulset.CreateStatefulSet(l.client, l.cluster.ID, namespace.Name, podTemplate, 3, true, l.longhornTestConfig.LonghornTestStorageClass)
+	require.NoError(l.T(), err)
+	l.T().Logf("Created StetefulSet %s on namespace %s", statefulSet.Name, namespaceName)
+
+	// The template we want will always be the last one on the list.
+	volumeSourceName := statefulSet.Spec.VolumeClaimTemplates[len(statefulSet.Spec.VolumeClaimTemplates)-1].Name
+	storage.CheckVolumeAllocation(l.T(), l.client, l.cluster.ID, namespace.Name, l.longhornTestConfig.LonghornTestStorageClass, volumeSourceName, storage.MountPath)
+
+	var stetefulSetPodReplicas int32 = 5
+	statefulSet.Spec.Replicas = &stetefulSetPodReplicas
+	statefulSet, err = statefulset.UpdateStatefulSet(l.client, l.cluster.ID, namespace.Name, statefulSet, true)
+	require.NoError(l.T(), err)
+
+	storage.CheckVolumeAllocation(l.T(), l.client, l.cluster.ID, namespace.Name, l.longhornTestConfig.LonghornTestStorageClass, volumeSourceName, storage.MountPath)
+
+	steveClient, err := l.client.Steve.ProxyDownstream(l.cluster.ID)
+	require.NoError(l.T(), err)
+
+	pvcBeforeScaling, err := steveClient.SteveType(persistentvolumeclaims.PersistentVolumeClaimType).NamespacedSteveClient(namespace.Name).List(nil)
+	require.NoError(l.T(), err)
+
+	stetefulSetPodReplicas = 2
+	statefulSet.Spec.Replicas = &stetefulSetPodReplicas
+	statefulSet, err = statefulset.UpdateStatefulSet(l.client, l.cluster.ID, namespace.Name, statefulSet, true)
+	require.NoError(l.T(), err)
+
+	l.T().Logf("Verifying old volumes still exist")
+	volumesAfterScaling, err := steveClient.SteveType(storage.PersistentVolumeEntityType).List(nil)
+	require.NoError(l.T(), err)
+	var volumeNamesAfterScaling []string
+	for _, volume := range volumesAfterScaling.Data {
+		volumeNamesAfterScaling = append(volumeNamesAfterScaling, volume.Name)
+	}
+
+	var pvcSpec corev1.PersistentVolumeClaimSpec
+	for _, pvc := range pvcBeforeScaling.Data {
+		err = steveV1.ConvertToK8sType(pvc.Spec, &pvcSpec)
+		require.NoError(l.T(), err)
+		require.True(l.T(), slices.Contains(volumeNamesAfterScaling, pvcSpec.VolumeName))
+	}
+
+	pods, err := steveClient.SteveType(shepherdPods.PodResourceSteveType).NamespacedSteveClient(namespace.Name).List(nil)
+	require.NoError(l.T(), err)
+	require.Equal(l.T(), 2, len(pods.Data))
+
+	err = steveClient.SteveType(shepherdPods.PodResourceSteveType).Delete(&pods.Data[0])
+	require.NoError(l.T(), err)
+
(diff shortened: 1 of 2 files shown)


actions/charts/longhorn.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return err
+	}
+
+	// TO uninstall Longhorn successfully it is first needed to set the "deleting-confirmation-flag" on Longhorn settings.
+	setDeletionFlagCommand := []string{"kubectl", "-n", namespace, "patch", "settings.longhorn.io", "deleting-confirmation-flag", "--type=merge", "-p", `'{"value": "true"}'`}
+	_, err = kubectl.Command(client, nil, clusterID, setDeletionFlagCommand, "")
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
hamistao: After a brief discussion with @slickwarren, we decided to leave this logic as part of the installation process as to not make this dependant on other PR over on shepherd. That said, the PR changing the behavior of creating a new session will be created and if/when merged, we can consider updating this to take advandage of this change.

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,264 @@
...
+	nginxSpec := &appv1.DeploymentSpec{}
+	err = steveV1.ConvertToK8sType(nginxResponse.Spec, nginxSpec)
+	require.NoError(l.T(), err)
+
+	// Even though the Longhorn default for number of replicas is 2, Rancher enforces its own default of 3.
+	volumeName := nginxSpec.Template.Spec.Volumes[0].Name
+	checkReplicasCommand := []string{"kubectl", "-n", longhornNamespace, "get", "volumes.longhorn.io", volumeName, `-o=jsonpath="{.spec.numberOfReplicas}"`}
+	settingValue, err := kubectl.Command(l.client, nil, l.cluster.ID, checkReplicasCommand, "")
+	require.NoError(l.T(), err)
+	require.Equal(l.T(), "\"3\"\n", settingValue)
```
hamistao: This behavior is very strange to me and I did not find anything explicitly stating that this is expected, but I assume it is the intended behavior and tested it as such. If you know of a better way to proceed here please let me know.

@rancher/qa-pit-crew

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,264 @@
...
+	if chart.IsAlreadyInstalled {
+		if l.installedLonghorn {
+			l.T().Log("Uninstalling Longhorn that was intalled on the previous test.")
+			err = charts.UninstallLonghornChart(l.client, longhornNamespace, l.cluster.ID, l.payloadOpts.Host)
+			require.NoError(l.T(), err)
+		} else {
+			l.T().Skip("Skipping installation test because Longhorn is already installed")
+		}
+	}
+
```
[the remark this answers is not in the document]
hamistao: cc @rancher/qa-pit-crew

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,264 @@
...
+	require.NoError(l.T(), err)
+
+	var restConfig *rest.Config
+	restConfig, err = (*kubeConfig).ClientConfig()
+	require.NoError(l.T(), err)
+
+	steveClient, err := l.client.Steve.ProxyDownstream(l.cluster.ID)
+	require.NoError(l.T(), err)
+
+	pods, err := steveClient.SteveType(pods.PodResourceSteveType).NamespacedSteveClient(namespaces.Default).List(nil)
```
lscalabrini01: Is it ok if pods return empty? should us add a require.NoEmpty condition?

validation/longhorn/longhorn_test.go
lscalabrini01: I think that there a better way to check the replicas, I'll let you know

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+func (l *LonghornTestSuite) TestChartInstallCustomConfig() {
+	chart, err := shepherdCharts.GetChartStatus(l.client, l.cluster.ID, longhornNamespace, longhornChartName)
+	require.NoError(l.T(), err)
+
+	// If Longhorn was installed by a previous test on this same session, uninstall it to install it again with custom configuration.
+	// If Longhorn was installed previously to this test run, leave it be and skip this test. This way we allow for running the
+	// next tests on top of a manually installed Longhorn and avoid accidentally uninstalling something important.
+	if chart.IsAlreadyInstalled {
+		if l.installedLonghorn {
```
lscalabrini01: I don’t think we need to check whether Longhorn is installed between tests, we can just run the tests with cleanup set to true

validation/longhorn/longhorn_test.go
lscalabrini01: Could we do something similar to what we have here? https://github.com/rancher/tests/blob/main/actions/workloads/deployment/verify.go#L31C6-L31C22

hamistao: Turns out this is expected after all, I can link you the thread where this was discussed later.

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,264 @@
...
+	require.NoError(l.T(), err)
+
+	debugPods, err := steveClient.SteveType(pods.PodResourceSteveType).NamespacedSteveClient(debugNamespace).List(nil)
+	require.NoError(l.T(), err)
+
+	podStatus := &corev1.PodStatus{}
+	err = steveV1.ConvertToK8sType(debugPods.Data[0].Status, podStatus)
+	require.NoError(l.T(), err)
+	require.Equal(l.T(), "Succeeded", string(podStatus.Phase))
+}
```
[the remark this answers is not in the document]
hamistao: Indeed this whole logic to check the node filesystem would look nice in a helper function, I will wait as you say to better decide where each function would go.

validation/longhorn/longhorn_test.go
hamistao: @lscalabrini01 please let me know what you think when you have the time

hamistao: Update: I am having trouble making this above work as it is, but this should show more or less how else we can do this.

actions/charts/longhorn.go
```diff
@@ -0,0 +1,137 @@
...
+
+	err = shepherdCharts.WatchAndWaitDeployments(client, payload.Cluster.ID, payload.Namespace, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	err = shepherdCharts.WatchAndWaitDaemonSets(client, payload.Cluster.ID, payload.Namespace, metav1.ListOptions{})
+	return err
```
floatingman: Is it possible to add label selectors that specifically target Longhorn workloads to the metav1.ListOptions instead of being blank? It's possible, though unlikely, that other workloads could be running in the namespace

actions/workloads/pods/pods.go
```diff
@@ -126,7 +125,7 @@ func CheckPodLogsForErrors(client *rancher.Client, clusterID string, podName str
 }
 
 // WatchAndWaitPodContainerRunning is a helper to watch and wait all pod containers running
-func WatchAndWaitPodContainerRunning(client *rancher.Client, clusterID, namespaceName string, deploymentTemplate *appv1.Deployment) error {
+func WatchAndWaitPodContainerRunning(client *rancher.Client, clusterID, namespaceName string) error {
```
floatingman: Did you ensure no other tests call this function with the removed parameter?

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+func (l *LonghornTestSuite) TestChartInstall() {
+	chart, err := shepherdCharts.GetChartStatus(l.client, l.cluster.ID, longhornNamespace, longhornChartName)
+	require.NoError(l.T(), err)
+
+	if chart.IsAlreadyInstalled {
+		l.T().Skip("Skipping installation test because Longhorn is already installed")
+	}
+
+	l.T().Logf("Installing Lonhgorn chart in cluster [%v] with latest version [%v] in project [%v] and namespace [%v]", l.cluster.Name, l.payloadOpts.Version, l.project.Name, l.payloadOpts.Namespace)
```
floatingman: "Lonhgorn" should be "Longhorn"

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,264 @@
...
+func (l *LonghornTestSuite) TestChartInstallCustomConfig() {
+	chart, err := shepherdCharts.GetChartStatus(l.client, l.cluster.ID, longhornNamespace, longhornChartName)
+	require.NoError(l.T(), err)
+
+	// If Longhorn was installed by a previous test on this same session, uninstall it to install it again with custom configuration.
+	// If Longhorn was installed previously to this test run, leave it be and skip this test. This way we allow for running the
+	// next tests on top of a manually installed Longhorn and avoid accidentally uninstalling something important.
+	if chart.IsAlreadyInstalled {
+		if l.installedLonghorn {
+			l.T().Log("Uninstalling Longhorn that was intalled on the previous test.")
```
floatingman: "intalled' should be "installed"

floatingman (changes requested):
Looks good overall. You should consider adding a README for the longhorn_test, so others will know how to run the tests and if any specific cattle-config settings are needed.

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,314 @@
+//go:build validation
```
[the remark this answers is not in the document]
rancher-max: yes we should!

rancher-max (approved):
LGTM -- left one comment that I feel is a nit so will leave up to you if you want to update or not

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,314 @@
...
+		l.T().Skip("Skipping installation test because Longhorn is already installed")
+	}
+
+	l.T().Logf("Installing Longhorn chart in cluster [%v] with latest version [%v] in project [%v] and namespace [%v]", l.cluster.Name, l.payloadOpts.Version, l.project.Name, l.payloadOpts.Namespace)
+	err = charts.InstallLonghornChart(l.client, l.payloadOpts, nil)
+	require.NoError(l.T(), err)
+	l.installedLonghorn = true
+}
+
+func (l *LonghornTestSuite) TestVolumeCreationOnWorkload() {
```
slickwarren: Typically with charts, we write the test for a given chart as follows:

* install the chart
* verify the chart is working correctly

in your tests, you split these two into different tests. Can you instead combine TestChartInstall and TestVolumeCreatonOnWorkoad into one, probably just TestChart or something similar?

actions/cloudprovider/verify.go
slickwarren: can you move this to a public const, so that it can be pulled from the test when you use it in longhorn_test.go? Instead of hardcoding it in both places.

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,314 @@
...
+	}})
+	require.NoError(l.T(), err)
+
+	// Label worker nodes to check effectiveness of createDefaultDiskLabeledNodes setting.
+	// Also save the name of one worker node for future use.
+	l.T().Log("Label worker nodes with Longhorn's create-default-disk=true")
+	var workerName string
+	for _, node := range nodeCollection.Data {
+		if node.Worker {
+			labelNodeCommand := []string{"kubectl", "label", "node", node.Hostname, "node.longhorn.io/create-default-disk=true"}
```
slickwarren: can you put the last string in a cost.

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,314 @@
...
+	require.NoError(l.T(), err)
+
+	steveClient, err := l.client.Steve.ProxyDownstream(l.cluster.ID)
+	require.NoError(l.T(), err)
+
+	l.session.RegisterCleanupFunc(func() error {
+		return steveClient.SteveType(namespaceActions.NamespaceSteveType).Delete(createdNamespace)
+	})
+
+	checkDataPathCommand := []string{"kubectl", "debug", "node/" + workerName, "-n", debugNamespace, "--profile=general", "--image=busybox", "--", "/bin/sh", "-c", "test -d /host/var/lib/longhorn-custom/replicas && test -f /host/var/lib/longhorn-custom/longhorn-disk.cfg"}
```
slickwarren: is there a reason you use kubectl instead of rancher api for stuff like this? Or in other words, is there not a way to check this via rancher instead of kubectl?

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	l.T().Logf("Write to mounted volume on pod [%v]", podName)
+	writeToMountedVolume := []string{"touch", "/auto-mnt/test-file"}
+	_, err = kubeconfig.KubectlExec(restConfig, podName, namespaces.Default, writeToMountedVolume)
+	require.NoError(l.T(), err)
+
+	checkFileOnVolume := []string{"stat", "/auto-mnt/test-file"}
+	_, err = kubeconfig.KubectlExec(restConfig, podName, namespaces.Default, checkFileOnVolume)
+	require.NoError(l.T(), err)
```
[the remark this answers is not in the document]
hamistao: The satefulset tests that I am working on right now does exactly what you are describing so maybe leaving this simpler approach here is OK

actions/cloudprovider/verify.go
```diff
@@ -300,6 +284,51 @@ func CreatePVCWorkload(t *testing.T, client *rancher.Client, clusterID string) *
 	return nginxResponse
 }
 
+// GetStorageClass gets a storage class with the provided name on the provided cluster.
+// If an empty storageClassName is provided, the first one on the list of available storage classes will be used.
```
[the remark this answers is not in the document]
hamistao: I can, I am not sure I should as I tried to preserve the original behavior of the function here and the way it used to work was to always get the first in the list. For my tests this has no impact since I am always providing a specific storage class anyway (longhorn), but it would affect the pre-existing ones.

scripts/capture_images/capture_images.go
```diff
@@ -107,31 +106,29 @@ func main() {
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, syscall.SIGTERM)
 
-	rancherConfig := new(rancher.Config)
-	config.LoadConfig(rancher.ConfigurationFileKey, rancherConfig)
-
-	client, err := rancher.NewClientForConfig("", rancherConfig, session.NewSession())
+	client, err := rancher.NewClient("", session.NewSession())
```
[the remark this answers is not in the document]
hamistao: I am the wrote `capture_images.go`, so when I found out that I could just use `NewClient` I thought "oh so I probably could have used it there as well" :D

actions/charts/longhorn.go
```diff
@@ -0,0 +1,171 @@
...
+			return false, fmt.Errorf("there was an error uninstalling Longhorn chart")
+		case watch.Deleted:
+			return true, nil
+		}
+		return false, nil
+	})
+}
+
+// This function has the purpose of helping to test RBAC with Longhorn.
+func CreateAndDeleteLonghornVolume(client *rancher.Client, clusterID string, namespace string, storageClass storagev1.StorageClass) error {
```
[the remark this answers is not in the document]
slickwarren: sgtm

actions/cloudprovider/verify.go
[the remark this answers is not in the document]
slickwarren: I think it would be best to update it if list[0] isn't guaranteed to be the default one. Can be a future PR if this was already the behavior before these changes.

validation/longhorn/longhorn_test.go
```diff
@@ -0,0 +1,350 @@
...
+	// The template we want will always be the last one on the list.
+	volumeSourceName := statefulSet.Spec.VolumeClaimTemplates[len(statefulSet.Spec.VolumeClaimTemplates)-1].Name
+	storage.CheckVolumeAllocation(l.T(), l.client, l.cluster.ID, namespace.Name, l.longhornTestConfig.LonghornTestStorageClass, volumeSourceName, storage.MountPath)
+
+	var stetefulSetPodReplicas int32 = 5
+	statefulSet.Spec.Replicas = &stetefulSetPodReplicas
+	statefulSet, err = statefulset.UpdateStatefulSet(l.client, l.cluster.ID, namespace.Name, statefulSet, true)
+	require.NoError(l.T(), err)
+
+	storage.CheckVolumeAllocation(l.T(), l.client, l.cluster.ID, namespace.Name, l.longhornTestConfig.LonghornTestStorageClass, volumeSourceName, storage.MountPath)
```
slickwarren: there's some weirdness in the logs around this function but I can't figure out exactly where:
```

09:09:24  2025/12/24 17:09:23 (0xc000efe000) (0xc000b600a0) Stream added, broadcasting: 3
09:09:24  2025/12/24 17:09:23 (0xc000efe000) Reply frame received for 3
09:09:24  2025/12/24 17:09:23 (0xc000efe000) (0xc000a663c0) Create stream
09:09:24  2025/12/24 17:09:23 (0xc000efe000) (0xc000a663c0) Stream added, broadcasting: 5
09:09:24  2025/12/24 17:09:23 (0xc000efe000) Reply frame received for 5
```

validation/longhorn/longhorn_test.go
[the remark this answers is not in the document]
slickwarren: 1 last nit here; put chart_install test in its own file -- move the rest out (same structure as validation/fleet/provisioning)

this way, the non-install tests can have the chart check + install / uninstall as a part of setup instead of in each test.

interoperability/longhorn/longhorn.go
```diff
@@ -0,0 +1,50 @@
...
+	defer func() {
+		if r := recover(); r != nil {
+			longhornTestConfig = TestConfig{
+				LonghornTestProject:      LonghornTestDefaultProject,
+				LonghornTestStorageClass: LonghornTestDefaultStorageClass,
+			}
+		} else {
+			if !slices.Contains(storage.LonghornStorageClasses, longhornTestConfig.LonghornTestStorageClass) {
+				panic(fmt.Errorf("Invalid storage class %s", longhornTestConfig.LonghornTestStorageClass))
+			}
```
slickwarren: won't this fail is its empty? Or maybe it never will be empty if you use the json suggestion above. . ?

interoperability/longhorn/longhorn.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+	}()
+
+	if longhornTestConfig.LonghornTestProject == "" {
+		longhornTestConfig.LonghornTestProject = LonghornTestDefaultProject
+	}
+
+	if longhornTestConfig.LonghornTestStorageClass == "" {
```
slickwarren: I think this will never be reached if it is indeed empty due to the panic above.

validation/longhorn/README.md
```diff
@@ -0,0 +1,15 @@
...
+This directory contains tests for interoperability between Rancher and Longhorn. The list of planned tests can be seen on `schemas/pit-schemas.yaml` and the implementation for the ones that are automated is contained in `longhorn_test.go`.
+
+## Running the tests
+
+To run all available tests, one just has to run the `TestLonghornTestSuite`. Additional configuration can be included in the Cattle Config file as follows:
+
+```yaml
+longhorn:
+  testProject: "longhorn-custom-test"
+  testStorageClass: "longhorn" # Can be either "longhorn" or "longhorn-static"
```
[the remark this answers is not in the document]
hamistao: tbh I am not sure. Since most operations perform over the course of the tests can be done with either, I figured I should add this option in case someone wants to test using `longhorn-static` specifically, but I am not sure as which scenarios would demand that.

interoperability/longhorn/longhorn.go
hamistao: This is for the case where the `longhorn` test configuration is there but one of the fields is empty.

interoperability/longhorn/longhorn.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/rancher/shepherd/pkg/config"
+	"github.com/rancher/tests/actions/storage"
+)
+
+// The json/yaml config key for the corral package to be build ..
+const (
+	longhornTestConfigConfigurationFileKey = "longhorn"
+	LonghornTestDefaultProject             = "longhorn-test"
+	LonghornTestDefaultStorageClass        = "longhorn"
+)
```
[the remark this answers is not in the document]
hamistao: Hey I like this approach. I am not sure whether it works if the configuration is missing entirely instead of just missing a few fields, but either way there is room for improvement here.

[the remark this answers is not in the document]
hamistao: Doing it this way was not possible, but instead I loaded the config into a struct populated with the default values, that has the same effect and greatly simplifies the code.

validation/longhorn/longhorn_test.go
floatingman: Seems like there might be a kubectl argument to change log verbosity. https://stackoverflow.com/questions/62660240/kubernetes-exec-return-the-mess-from-container

PR #415, opened 2025-12-10
Fixes to harvester job

slickwarren:
harvester job was breaking, this fixes it;

* updates to dev-provider.sh to allow a specific commit/branch version (fix for issue on harvester provider)
* fix for how dev version is found/applied/built

### DIFF
--- a/validation/harvester/import_harvester_test.go
+++ b/validation/harvester/import_harvester_test.go
@@ -1,4 +1,4 @@
-//go:build validation
+//go:build validation || harvester
 
 package harvester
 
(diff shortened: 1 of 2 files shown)


validation/harvester/import_harvester_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build validation
+//go:build validation || pit.daily
```
rancher-max: will having this one in pit.daily cause any weird issues? Like, will it try to run just this test but not have a Harvester environment built to actually use?

lscalabrini01: May be we can add two tags `pit.daily` and  `harvester`

rancher-max (approved):

PR #416, opened 2025-12-10
workload reworks

susesgartner:
Description: Updates the workload tests into reusable validations that can be used in most of our test cases. For now i've added these validations to a few of our provisioning tests, assuming it causes no flakiness I will add it to more tests. 

Changes:
* Split the createVerify functions into seperate create/verify functions (use cases snapshot, upgrade, autoscaler ... etc)
* Updated the create logic to accept the upstream objects so that our tests can fully configure the workloads.
* Modified some of the verifies to have more robust checks
* Updated the logging to support log filtering and have cleaner log messages.
* Updated the clusterPod verify to no longer use testing t so it can be used in other validations.


navsalehi (approved):

rancher-max (approved):
Should be fine from PIT perspective. Thank you for updating the schemas!

PR #433, opened 2025-12-23
[Qase] Make necessary changes to transition to the latest qase packages

### LINKED ISSUE #432: Qase packages in-use are outdated
We are using the old `go.qase.io` urls for our qase packages and using `replace` directives to bandaid over them. This is leading to some golang dependency issues for a big overhaul to `rancher/dartboard` which now adds dependencies on `rancher/shepherd` and `rancher/tests` in order to leverage existing code.

Barring a known issue that I am not aware of, we should be able to safely transition to the new and improved qase packages without losing any functionality. This would solve the dependency issue happening downstream 🙏

git-ival:
Closes #432 

Note, I have not tested this with any `rancher/tests` code that relies on the qase-related pieces since I am unsure how to go about doing that. If anyone could point me to what would be considered an acceptable smoke test and how to run them, then I'm happy to report back once they're done.


rancher-max:
I can point you to what to run on the PIT side! You should be able to run `go-pit-daily-job`, pointing it to your repo/branch. This should cover most things for all teams, but I will let others chime in too because I believe there are some differences.

PR #438, opened 2026-01-06
Update validation logic

susesgartner:
Modularize verify logic

Previously we added a check to the VerifyClusterPods that verifies certain required deployments spin up (SUC, fleet, rancher-webhook, cluster-agent). 

Changes:
* Updated pod/deployment logic to be modular.
* Update deployment logic to have clear error messages instead of "context deadline exxceeded"
* Updated tests to match the new structure.

### DIFF
--- a/actions/workloads/deployment/verify.go
+++ b/actions/workloads/deployment/verify.go
@@ -4,12 +4,17 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strconv"
 	"time"
 
 	"github.com/rancher/shepherd/clients/rancher"
+	steveV1 "github.com/rancher/shepherd/clients/rancher/v1"
+	v1 "github.com/rancher/shepherd/clients/rancher/v1"
 	"github.com/rancher/shepherd/extensions/charts"
+	"github.com/rancher/shepherd/extensions/clusters"
 	"github.com/rancher/shepherd/extensions/defaults"
+	"github.com/rancher/shepherd/extensions/defaults/stevetypes"
 	"github.com/rancher/shepherd/extensions/workloads"
 	namegen "github.com/rancher/shepherd/pkg/namegenerator"
 	"github.com/rancher/shepherd/pkg/wrangler"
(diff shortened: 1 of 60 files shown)


navsalehi (approved):

actions/workloads/deployment/verify.go
```diff
@@ -490,3 +499,89 @@ func VerifyDeploymentOrchestration(client *rancher.Client, clusterID, namespace,
...
+
+			if slices.Contains(requiredDeployments, k8sDeployment.Name) {
+				requiredDeployments = slices.Delete(requiredDeployments, slices.Index(requiredDeployments, k8sDeployment.Name), slices.Index(requiredDeployments, k8sDeployment.Name)+1)
+			}
+		}
+		if len(requiredDeployments) != 0 {
+			return false, nil
+		}
+
+		return true, nil
```
slickwarren: there's never an error returned here. just confirming that, the only error you care about is a timeout? If so can yo uchange line 547 to something like "timeout waiting on required deployments.."?

actions/workloads/deployment/verify.go
```diff
@@ -490,3 +499,89 @@ func VerifyDeploymentOrchestration(client *rancher.Client, clusterID, namespace,
...
+			err := steveV1.ConvertToK8sType(deploymentObj.JSONResp, k8sDeployment)
+			if err != nil {
+				return false, nil
+			}
+
+			if k8sDeployment.Status.AvailableReplicas != *k8sDeployment.Spec.Replicas {
+				failedDeployments = append(failedDeployments, *k8sDeployment)
+				return false, nil
+			}
+		}
```
slickwarren:
```suggestion
			if k8sDeployment.Status.AvailableReplicas != *k8sDeployment.Spec.Replicas {
				failedDeployments = append(failedDeployments, *k8sDeployment)
				return false, nil
			}
		}
		if len(failedDeployments) > 0{
		    return false, Error.new("some deployments have failed validation...")
		}
		...
```
current version will always be at most len(1) afaik

markusewalker (approved):

PR #447, opened 2026-01-12
CAPI interoperability tests

lscalabrini01:


actions/charts/neuvector.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = waitUninstallation(catalogClient, namespace, neuVectorChartName)
+	if err != nil {
+		return err
+	}
+
+	if neuVectorChartName == NeuVectorChartName {
```
slickwarren: when is it possible this isn't true? In fact I would expect this to be hard-coded, right?

validation/charts/neuvector_test.go
```diff
@@ -0,0 +1,96 @@
...
+					Version:   latestNeuVectorChartVersion,
+					ProjectID: i.project.ID,
+				},
+			}
+
+			i.T().Logf("Setting up %s on cluster (%s)", neuVectorChartName, i.cluster.Name)
+			err = charts.InstallNeuVectorChart(i.client, neuVectorChartName, payloadOpts)
+			require.NoError(i.T(), err)
+		})
+	}
```
slickwarren: is there some validation you can add on this besides checking that the chart itself is active? i.e. check all the workloads from the neuvector namespace, or do a specific deployment on the cluster that uses neuvector chart?

validation/charts/cis_benchmark_test.go
```diff
@@ -0,0 +1,84 @@
...
+	chartInstallOptions := &charts.InstallOptions{
+		Cluster:   clusterMeta,
+		Version:   latestHardenedChartVersion,
+		ProjectID: i.project.ID,
+	}
+
+	i.T().Logf("Setting up %s on cluster (%s)", chartName, i.cluster.Name)
+	err = cis.SetupHardenedChart(i.client, i.project.ClusterID, chartInstallOptions, chartName, chartNamespace)
+	require.NoError(i.T(), err)
+}
```
slickwarren: is there some validation you can include after this step to ensure the chart is working properly, not just that it installed?

slickwarren (changes requested):
main points are:
* redundant naming in test name vs. suite name
* add validation after the chart(s) are installed
* probably move stuff to their own folder in validations/ like what was done for longhorn
* probably move neuvector shared functions from actions/ -> interoperability

actions/charts/neuvector.go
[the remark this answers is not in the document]
lscalabrini01: Yes, we should, to move it some refactors are need, the longhorn and the neuvector tests are using some actions structures on the tests

validation/charts/cis_benchmark_test.go
```diff
@@ -0,0 +1,84 @@
...
+	i.cluster = cluster
+
+	clusterMeta, err := extensionscluster.NewClusterMeta(i.client, i.cluster.Name)
+	require.NoError(i.T(), err)
+
+	i.project, err = projects.GetProjectByName(i.client, clusterMeta.ID, cis.System)
+	require.NoError(i.T(), err)
+}
+
+func (i *CisBenchmarkTestSuite) TestInstallCisBenchmarkChart() {
```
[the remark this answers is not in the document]
lscalabrini01: done!

validation/charts/neuvector_test.go
lscalabrini01: Yes, there is, and we are already checking it, NeuVector charts creates deployments and deamonset when it is installed and the InstallNeuVectorChart function is checking it

```
err = shepherdCharts.WatchAndWaitDeployments(client, payload.Cluster.ID, payload.Namespace, metav1.ListOptions{})
	if err != nil {
		return err
	}

err = shepherdCharts.WatchAndWaitDaemonSets(client, payload.Cluster.ID, payload.Namespace, metav1.ListOptions{})
	return err`
```

hamistao (commented):
The main comment here would be the big one on reworking InstallNeuVectorChart

hamistao (approved):
Thanks again for the help with Longhorn!!

actions/charts/charts.go
```diff
@@ -10,6 +10,7 @@ const (
 	rancherChartsName        = "rancher-charts"
 	rancherPartnerCharts     = "rancher-partner-charts"
 	active                   = "active"
+	SystemProject            = "System"
```
susesgartner: Nit: I believe this is only used in this package so it should be kept lower case.

susesgartner (dismissed):
LGTM

actions/charts/rancheralerting.go
```diff
@@ -84,59 +81,6 @@ func InstallRancherAlertingChart(client *rancher.Client, installOptions *Install
...
-				if event.Type == watch.Deleted {
-					return true, nil
-				}
-				return false, nil
-			})
-			if err != nil {
-				return err
-			}
-		}
-
```
floatingman: Why did all this get removed?

actions/charts/ranchermonitoring.go
```diff
@@ -114,51 +112,7 @@ func InstallRancherMonitoringChart(client *rancher.Client, installOptions *Insta
...
-		if err != nil {
-			return err
-		}
-
-		return wait.WatchWait(watchNamespaceInterface, func(event watch.Event) (ready bool, err error) {
-			if event.Type == watch.Deleted {
-				return true, nil
-			}
-			return false, nil
-		})
```
floatingman: Why was this removed?

validation/charts/alerting_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	clusterMeta, err := extensionscluster.NewClusterMeta(i.client, i.cluster.Name)
+	require.NoError(i.T(), err)
+
+	i.project, err = projects.GetProjectByName(i.client, clusterMeta.ID, charts.SystemProject)
+	require.NoError(i.T(), err)
+	require.Equal(i.T(), charts.SystemProject, i.project.Name)
+}
+
+func (i *AlertingTestSuite) TestAlertingInstallation() {
```
floatingman: At some point, we should add more extensive tests beyond just checking that the installation was successful.
Some ideas:
Verify alerting functionality
Test SMS/Teams drivers if enabled
Verify integration with monitoring

validation/charts/logging_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	i.T().Logf("Creating project [%s] on cluster [%s]", charts.SystemProject, cluster.Name)
+	createdProject, err := client.Management.Project.Create(projectConfig)
+	require.NoError(i.T(), err)
+	require.Equal(i.T(), charts.SystemProject, createdProject.Name)
+
+	i.project = createdProject
+}
+
+func (i *LoggingTestSuite) TestLoggingInstallation() {
```
floatingman: At some point, we should add more extensive tests beyond just checking that the installation was successful.

Some ideas:
Verify logs are being collected
Test AdditionalLoggingSources if enabled
Verify logging pipeline is working

validation/charts/monitoring_test.go
```diff
@@ -184,13 +188,22 @@ func (m *MonitoringTestSuite) TestMonitoringChart() {
...
-	for _, node := range nodeCollection.Data {
-		workerNodePublicIPs = append(workerNodePublicIPs, node.Annotations["rke.cattle.io/external-ip"])
+	for _, machine := range nodeList.Data {
+		logrus.Info("Getting the node IP")
+		newNode := &corev1.Node{}
+		err = steveV1.ConvertToK8sType(machine.JSONResp, newNode)
+		require.NoError(m.T(), err)
+
+		nodeIP := kubeapinodes.GetNodeIP(newNode, corev1.NodeExternalIP)
+		workerNodePublicIPs = append(workerNodePublicIPs, nodeIP)
```
floatingman: This is good, but one suggestion.


```suggestion
  for _, machine := range nodeList.Data {
       newNode := &corev1.Node{}
       err = steveV1.ConvertToK8sType(machine.JSONResp, newNode)
       if err != nil {
           logrus.Warnf("Failed to convert node: %v", err)
           continue  // Skip invalid nodes instead of failing completely
       }
  
       nodeIP := kubeapinodes.GetNodeIP(newNode, corev1.NodeExternalIP)
       if nodeIP == "" {
           logrus.Warnf("Node %s has no external IP", newNode.Name)
           continue
       }
       workerNodePublicIPs = append(workerNodePublicIPs, nodeIP)
   }
  
   require.NotEmpty(m.T(), workerNodePublicIPs, "No worker nodes with external IPs found")
```
If converting the type fails, a log entry will be created, and the final check will produce the same result as your function.

validation/charts/alerting_test.go
floatingman: Also, TestAlertingInstallation is redundant with the suite name. Just needs to be TestInstallation

validation/charts/cis_benchmark_test.go
```diff
@@ -0,0 +1,86 @@
+//go:build (validation || infra.any || i.cluster.any || sanity || pit.daily) && !stress && !extended
```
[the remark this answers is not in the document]
lscalabrini01: done

validation/charts/alerting_test.go
lscalabrini01: For while, we need to keep the redundant names to qase report be able to match the test and test case, I'll investigate why it is not considering the test suite name anymore and redundant test names can be refactor in the future

actions/charts/rancheralerting.go
lscalabrini01: It was removed because the tests were propagating the cleanup logic and attempting to delete the namespaces twice

validation/snapshot/rke2/schemas/pit_schemas.yaml
```diff
@@ -0,0 +1,48 @@
+- projects: [RANCHERINT]
```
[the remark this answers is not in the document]
lscalabrini01: We definitely want to add the schema file and the snapshot build tags for k3s in the future, for now, however, the RKE2 snapshot coverage is sufficient

validation/charts/alerting_test.go
lscalabrini01: I agree, I'll create a ticket to improve logging, alerting and monitoring charts

floatingman (approved):

validation/charts/cis_benchmark_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	clusterName := client.RancherConfig.ClusterName
+	require.NotEmptyf(c.T(), clusterName, "Cluster name to install is not set")
+
+	cluster, err := clusters.NewClusterMeta(client, clusterName)
+	require.NoError(c.T(), err)
+	c.cluster = cluster
+
+	clusterMeta, err := extensionscluster.NewClusterMeta(c.client, c.cluster.Name)
+	require.NoError(c.T(), err)
```
hamistao: These two function calls do the same thing and the same package is being imported with two different names.

validation/charts/cis_benchmark_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	c.T().Logf("Creating Project [%s]", cis.System)
+	c.project, err = projects.GetProjectByName(c.client, clusterMeta.ID, cis.System)
+	require.NoError(c.T(), err)
+	require.Equal(c.T(), cis.System, c.project.Name)
```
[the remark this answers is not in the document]
lscalabrini01: done!

validation/charts/alerting_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	clusterName := client.RancherConfig.ClusterName
+	require.NotEmptyf(a.T(), clusterName, "Cluster name to install is not set")
+
+	cluster, err := clusters.NewClusterMeta(client, clusterName)
+	require.NoError(a.T(), err)
+	a.cluster = cluster
+
+	clusterMeta, err := extensionscluster.NewClusterMeta(a.client, a.cluster.Name)
+	require.NoError(a.T(), err)
```
hamistao: There is the same problem regarding a duplicate call to NewClusterMeta here

PR #451, opened 2026-01-12
nightly build for harvester job

slickwarren:


validation/pipeline/tfp/Jenkinsfile.harvester.e2e
```diff
@@ -219,6 +217,10 @@ node("harvester-vpn-1") {
                         sh returnStdout: true, script: "echo 'finished'"
 
                         config = sh(script: "cat ${filename}", returnStdout: true).trim()
+
+                        sh '''#!/bin/bash
+                        cat config.yml
+                        '''
```
rancher-max: I think this is fine, but would it work the same if it were just `sh cat config.yml`?

slickwarren: yep, that can be rm'd 😅

PR #486, opened 2026-01-29
Fix flakiness for custom cluster tests

susesgartner:
Currently the custom cluster tests occasionally fail with a context deadline exceeded error. This can also occur in all custom cluster tests across the framework

Changes:
* Updated the VerifyClusterReady func to return an error to the user 
Note: this brings it inline with our other verifications and allows it to be reused in funcs without requiring testing T. In this case I added it to the custom cluster provisioning since its not flakey in comparison to the watch func.
* Updated the Timeout for VerifyDeployment to handle windows deployments.
* Updated the logging around VerifyDeployment to report the last known status when a context deadline is exceeded.

While this PR does help fix custom cluster flakiness and make flakiness in windows tests less frequent it will not solve all snapshot restore flakiness as there are other deep rooted issues that are still present in the snapshot package.

### DIFF
--- a/actions/etcdsnapshot/creates.go
+++ b/actions/etcdsnapshot/creates.go
@@ -23,9 +23,12 @@ import (
 	"github.com/rancher/shepherd/extensions/workloads/pods"
 	namegen "github.com/rancher/shepherd/pkg/namegenerator"
 	"github.com/rancher/tests/actions/config/defaults"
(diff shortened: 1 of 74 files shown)


navsalehi (dismissed):

Josh-Diamond (dismissed):

slickwarren (commented):
lgtm for pit crew 👍🏼

PR #499, opened 2026-02-04
Update verify deleted nodes func to use steve instead of norman

Josh-Diamond:
This PR aims to update our verify deleted nodes function, by moving away from norman, in favor of steve, which resolves the `context deadline exceeded` error which stemmed from that function.

### DIFF
--- a/actions/provisioning/verify.go
+++ b/actions/provisioning/verify.go
@@ -303,8 +303,33 @@ func VerifyDeleteRKE2K3SCluster(t *testing.T, client *rancher.Client, clusterID 
 	require.NoError(t, err)
 
 	logrus.Infof("Waiting for nodes to be deleted on cluster (%s)", clusterID)
-	err = nodestat.AllNodeDeleted(client, clusterID)
+	err = VerifyAllNodesDeleted(client, clusterID)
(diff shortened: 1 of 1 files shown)


actions/provisioning/verify.go
```diff
@@ -303,10 +303,35 @@ func VerifyDeleteRKE2K3SCluster(t *testing.T, client *rancher.Client, clusterID
 	require.NoError(t, err)
 
 	logrus.Infof("Waiting for nodes to be deleted on cluster (%s)", clusterID)
-	err = nodestat.AllNodeDeleted(client, clusterID)
+	err = VerifyAllNodesDeleted(t, client, clusterID)
 	require.NoError(t, err)
 }
 
+// VerifyAllNodesDeleted validates that a non-rke1 cluster's nodes have been successfully deleted.
+func VerifyAllNodesDeleted(t *testing.T, client *rancher.Client, clusterID string) error {
```
susesgartner: For the sake of re usability and to keep consistent with all other verification functions. Please remove testing.T from this function

susesgartner: Please remove testing.T from this since it is unnecessary

PR #500, opened 2026-02-05
Check tests failing on QASE

lscalabrini01:


interoperability/fleet/fleet.go
```diff
@@ -143,7 +148,8 @@ func VerifyGitRepo(client *rancher.Client, gitRepoID, k8sClusterID, steveCluster
 			return true, errors.New(gitRepo.State.Message)
 		}
 
-		if gitStatus.Summary.NotReady == 0 && gitStatus.ReadyClusters == gitStatus.Summary.DesiredReady {
+		// Added a condition to handle cases where the fleet remains in a Modified state even though all resources are ready
```
slickwarren: this also sounds like a bug, no?

interoperability/fleet/fleet.go
```diff
@@ -143,7 +148,8 @@ func VerifyGitRepo(client *rancher.Client, gitRepoID, k8sClusterID, steveCluster
...
+		if (gitStatus.Summary.NotReady == 0 && gitStatus.ReadyClusters == gitStatus.Summary.DesiredReady) || (gitStatus.Summary.NotReady == 0 && gitStatus.Summary.Modified > 0 && gitStatus.ResourceCounts.NotReady == 0) {
```
slickwarren: `gitStatus.Summary.NotReady == 0 && gitStatus.ReadyClusters == gitStatus.Summary.DesiredReady` is different than the commented reason. does this need to be in?

lscalabrini01: The comment is about the second condition `gitStatus.Summary.NotReady == 0 && gitStatus.Summary.Modified > 0 && gitStatus.ResourceCounts.NotReady == 0,` I'll update the comment to include the first condition

interoperability/fleet/fleet.go
lscalabrini01: yes, it a bug, using the fleet:109.0.0+up0.15.0-alpha.4 version, it works fine

actions/charts/ranchermonitoring.go
```diff
@@ -329,3 +330,21 @@ func addMonitoringProviderPrefix(provider clusters.KubernetesProvider, opts *Ran
 
 	return newOptsMap, nil
 }
+
+// DeleteMonitoringResources follows the Monitoring uninstall reference guide
+// and deletes the ValidatingWebhookConfiguration to avoid installation conflicts
+// added the --ignore-not-found=true flag and xargs logic to prevent “not found” errors
+// Doc: There is no documentation available as a reference, but the information is described in the chart
+func DeleteMonitoringResources(client *rancher.Client, clusterID string) (string, error) {
+	deleteCommand := []string{
```
floatingman: Add a logrus for these cleanup functions

floatingman (changes requested):
Just a few suggestions and one critical thing to take care of.

actions/charts/ranchermonitoring.go
```diff
@@ -329,3 +330,21 @@ func addMonitoringProviderPrefix(provider clusters.KubernetesProvider, opts *Ran
...
+		"sh", "-c",
+		fmt.Sprintf("%s && %s && %s && %s && %s",
+			"kubectl get crds -oname | grep 'monitoring.coreos.com' | xargs kubectl delete",
+			"kubectl delete mutatingwebhookconfiguration rancher-monitoring-admission --ignore-not-found=true",
+			"kubectl delete validatingwebhookconfiguration rancher-monitoring-admission --ignore-not-found=true",
+			"kubectl get CustomResourceDefinition -l='app.kubernetes.io/part-of=istio' -o name -A | xargs -r kubectl delete",
+			"kubectl delete apiservices v1beta1.custom.metrics.k8s.io --ignore-not-found=true"),
```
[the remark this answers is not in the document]
lscalabrini01: I am afraid to use a loop because every time the command runs, it creates a new job and pod to execute it, if it fails, it will retry four times, and the loop would also increases the test duration, additionally, we would need to handle errors properly, if one of the commands fails, we should continue running the remaining commands and only return the aggregated error at the end

PR #504, opened 2026-02-06
Rework scaling package

susesgartner:
Changes:
* Removed the custom scaling test due to it being redundant.
* Moved the replace test to the deleting package since its more of a deleting test then a scaling test.
* Updated a variety of functions to remove flakiness and improve re-usability.
* Add existing cluster support


actions/machines/verify.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/rancher/shepherd/clients/rancher"
+	v1 "github.com/rancher/shepherd/clients/rancher/v1"
+	"github.com/rancher/shepherd/extensions/defaults"
+	"github.com/rancher/shepherd/extensions/defaults/stevetypes"
+	"github.com/sirupsen/logrus"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+func VerifyMachineReplacement(client *rancher.Client, replacedMachine *v1.SteveAPIObject) error {
```
Josh-Diamond: Function name is self-explanatory IMO, but our other helper functions have a comment briefly describing the behavior/use case of the function.  I'll still approve this PR, but wanted to mention I noticed this one didn't have a comment 🙂  Small change if we would like that.

PR #549, opened 2026-03-06
Upgrade and validate rancher within Harvester

lscalabrini01:


slickwarren (approved):
lgtm but why are there 7 commits for this?

PR #552, opened 2026-03-09
add qa-infra-automation as a provider via interoperability

slickwarren:
This PR connects rancher/tests -> rancher/qa-infra-automation by adding it as a client to rancher/tests/interoperability. 

Feedback on design is welcome. Overall, it looks like this:

* clone rancher/qa-infra-automation (An alternative to this is, we could release versions and put it in interoperability/go.mod)
* add qaInfraAutomation to your config (see interoperability/README.md for examples / placeholders)
* add a provisioning option to your test (see validation/neuvector/neuvector-hardened-test.go for an example)

in principle, this should only be used as a 'setup' option, when a test needs access to a downstream cluster to do some tests. 

leaving the neuvector test in for now to provide an example, but for complexity's sake I can put this in a separate PR upon request.

### DIFF
--- a/interoperability/qainfraautomation/tofu/tofu.go
+++ b/interoperability/qainfraautomation/tofu/tofu.go
@@ -36,7 +36,7 @@ func run(name string, args []string, dir string, env []string) ([]byte, error) {
 	cmd := exec.Command(name, args...)
 	cmd.Dir = dir
 	if len(env) > 0 {
-		cmd.Env = env
+		cmd.Env = append(cmd.Env, env...)
 	}
 	out, err := cmd.CombinedOutput()
 	if err != nil {
--- a/actions/uiplugins/uiplugins.go
+++ b/actions/uiplugins/uiplugins.go
@@ -106,7 +106,7 @@ func InstallUIPlugin(client *rancher.Client, installExtensionOptions *ExtensionO
 		}
 	}
 
-	return err
+	return nil
 }
 
 // CreateExtensionsRepo is a helper that utilizes the rancher client and add the ui extensions repo to the list if repositories in the local cluster.


interoperability/qainfraautomation/provisioning.go
```diff
@@ -0,0 +1,701 @@
...
+	data, err := json.MarshalIndent(v, "", "  ")
+	if err != nil {
+		return "", fmt.Errorf("marshal tfvars JSON: %w", err)
+	}
+	destPath := filepath.Join(repoPath, filename)
+	if err := os.WriteFile(destPath, data, 0600); err != nil {
+		return "", fmt.Errorf("write tfvars file %s: %w", destPath, err)
+	}
+	logrus.Debugf("[qainfraautomation] wrote tfvars: %s", destPath)
+	return destPath, nil
```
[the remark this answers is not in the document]
slickwarren: I think this could be addressed in a separate PR if it becomes important.

interoperability/qainfraautomation/ansible/ansible.go
```diff
@@ -0,0 +1,146 @@
...
+	out, err := c.run("ssh-add", []string{privateKeyPath}, c.repoPath, nil)
+	if err != nil {
+		return fmt.Errorf("ssh-add: %w", err)
+	}
+	logrus.Debugf("[ansible] ssh-add output:\n%s", string(out))
+	return nil
+}
+
+// GenerateInventory renders templatePath with env substitutions and writes the result to outputPath.
+func (c *Client) GenerateInventory(templatePath, outputPath string, env map[string]string) error {
```
hamistao: I think generally we should generate the inventory using tofu as we do for airgap. This seems like a step in the wrong direction and instead we should enable inventory generation for non-airgap. I would volunteer to do it if we agree that this is the way to go.

interoperability/qainfraautomation/tofu/tofu.go
```diff
@@ -0,0 +1,200 @@
...
+	Resources []ShowResource `json:"resources"`
+}
+
+// ShowResource represents a single resource entry from `tofu show -json`.
+type ShowResource struct {
+	Address string                     `json:"address"`
+	Type    string                     `json:"type"`
+	Name    string                     `json:"name"`
+	Values  map[string]json.RawMessage `json:"values"`
+}
```
hamistao: Maybe move these types to the top of the file.

interoperability/qainfraautomation/README.md
```diff
@@ -0,0 +1,136 @@
...
+    volumeType: gp3
+    hostnamePrefix: prefix
+    route53Zone: example.domain
+    airgapSetup: false
+    proxySetup: false
+    nodes:
+      - count: 1
+        role: [etcd, cp]
+      - count: 2
+        role: [worker]
```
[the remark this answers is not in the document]
slickwarren: would updating the name to something better make your answer change? i.e. in provisioiningInput we have MachinePools and NodePools instead of just node.

actions/charts/neuvector_new.go
floatingman: Why is this named with _new?

interoperability/qainfraautomation/provisioning.go
```diff
@@ -0,0 +1,850 @@
...
+		t.Fatalf("ssh-add: %v", err)
+	}
+
+	playbookEnv := []string{
+		"TF_WORKSPACE=" + workspace,
+		"TERRAFORM_NODE_SOURCE=" + nodeModulePath,
+		"RANCHER_URL=https://" + rancherClient.RancherConfig.Host,
+		"RANCHER_TOKEN=" + rancherClient.RancherConfig.AdminToken,
+	}
+	if *rancherClient.RancherConfig.Insecure {
```
floatingman: This needs a nil check in case Insecure is nil for some reason

interoperability/qainfraautomation/provisioning.go
```diff
@@ -0,0 +1,850 @@
...
+		t.Fatalf("unsupported cluster type: %s (must be rke2 or k3s)", clusterType)
+	}
+
+	destKubeconfig := filepath.Join(repoPath, harvesterKubeconfigDest)
+	logrus.Infof("[qainfraautomation] copying Harvester kubeconfig %s → %s", h.KubeConfigPath, destKubeconfig)
+	if err := copyFile(h.KubeConfigPath, destKubeconfig); err != nil {
+		t.Fatalf("copy harvester kubeconfig: %v", err)
+	}
+
+	vmVars := buildHarvesterVMVars(h, cfg.StandaloneCluster.Nodes, sshPubKeyContents)
```
floatingman: Why isn't this using clusterCfg.Nodes since you declared it above?

floatingman (commented):
Just a few suggestions.

PR #570, opened 2026-03-20
Automation for SCIM endpoints

dasarinaidu:
1. Automated SCIM endpoints
2. Started with OpenLdap AuthProvider
3. Will add more provider tests in upcoming PRs

### DIFF
--- a/actions/scim/scim.go
+++ b/actions/scim/scim.go
@@ -1,0 +1,262 @@
+package scim
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"github.com/rancher/shepherd/clients/rancher"
+	scimclient "github.com/rancher/shepherd/clients/rancher/auth/scim"
+	"github.com/rancher/shepherd/extensions/defaults"
+	"github.com/rancher/shepherd/pkg/clientbase"
+	namegen "github.com/rancher/shepherd/pkg/namegenerator"
+	"github.com/rancher/tests/actions/auth"
+	"github.com/rancher/tests/actions/features"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	SCIMFeatureFlag     = "scim"
+	SCIMSecretNamespace = "cattle-global-data"
+	SCIMSecretDataKey   = "token"
+
+	scimSecretKindLabel   = "cattle.io/kind"
+	scimSecretKindValue   = "scim-auth-token"
+	scimAuthProviderLabel = "authn.management.cattle.io/provider"
+)
+
+var errNoSCIMTokenSecret = fmt.Errorf("no SCIM token secret found")
+
+// FetchSCIMBearerToken retrieves the SCIM bearer token for the given auth provider from the cattle-global-data namespace
+func FetchSCIMBearerToken(client *rancher.Client, providerName string) (string, error) {
+	logrus.Infof("Fetching SCIM bearer token from %s by label %s=%s",
+		SCIMSecretNamespace, scimAuthProviderLabel, providerName)
+	selector := labels.SelectorFromSet(labels.Set{
+		scimSecretKindLabel:   scimSecretKindValue,
+		scimAuthProviderLabel: providerName,
+	})
+	list, err := client.WranglerContext.Core.Secret().List(
+		SCIMSecretNamespace,
+		metav1.ListOptions{LabelSelector: selector.String()},
+	)
+	if err != nil {
+		return "", err
+	}
+	if len(list.Items) == 0 {
+		return "", errNoSCIMTokenSecret
+	}
+	newest := &list.Items[0]
+	for i := 1; i < len(list.Items); i++ {
+		if list.Items[i].CreationTimestamp.After(newest.CreationTimestamp.Time) {
+			newest = &list.Items[i]
+		}
+	}
+	if len(list.Items) > 1 {
+		logrus.Warnf("Multiple SCIM token secrets found for provider %s, using newest: %s", providerName, newest.Name)
+	}
+	token, ok := newest.Data[SCIMSecretDataKey]
+	if !ok || len(token) == 0 {
+		return "", fmt.Errorf("key %q not found or empty in secret %s/%s",
+			SCIMSecretDataKey, SCIMSecretNamespace, newest.Name)
+	}
+	logrus.Infof("Found SCIM token in secret %s/%s", SCIMSecretNamespace, newest.Name)
+	return string(token), nil
+}
+
+// CreateSCIMTokenSecret generates a new random bearer token and stores it as a Kubernetes secret in the cattle-global-data namespace
+func CreateSCIMTokenSecret(client *rancher.Client, providerName string) (string, error) {
+	b := make([]byte, 32)
+	if _, err := rand.Read(b); err != nil {
+		return "", err
+	}
+	token := hex.EncodeToString(b)
+
+	secretName := fmt.Sprintf("scim-token-%s", providerName)
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: SCIMSecretNamespace,
+			Labels: map[string]string{
+				scimSecretKindLabel:   scimSecretKindValue,
+				scimAuthProviderLabel: providerName,
+			},
+		},
+		Data: map[string][]byte{
+			SCIMSecretDataKey: []byte(token),
+		},
+	}
+
+	logrus.Infof("Creating SCIM token secret %s/%s", SCIMSecretNamespace, secretName)
+	_, err := client.WranglerContext.Core.Secret().Create(secret)
+	if err != nil {
+		return "", err
+	}
+	return token, nil
+}
+
+// SetupSCIMClient ensures the SCIM feature flag is enabled and the auth provider
+// is active, then fetches or creates a bearer token secret and returns a configured
+// SCIM client ready to make requests against the given provider.
+func SetupSCIMClient(client *rancher.Client, providerName string) (*scimclient.Client, error) {
+	enabled, err := features.IsEnabled(client, SCIMFeatureFlag)
+	if err != nil {
+		return nil, err
+	}
+
+	if !enabled {
+		if err = features.UpdateFeatureFlag(client, SCIMFeatureFlag, true); err != nil {
+			return nil, err
+		}
+	}
+
+	if err = auth.EnsureAuthProviderEnabled(client, providerName); err != nil {
+		return nil, err
+	}
+
+	token, err := FetchSCIMBearerToken(client, providerName)
+	if err != nil {
+		if err != errNoSCIMTokenSecret {
+			return nil, err
+		}
+		logrus.Infof("No SCIM token secret found for provider %s, creating one", providerName)
+		token, err = CreateSCIMTokenSecret(client, providerName)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return scimclient.NewClient(&clientbase.ClientOpts{
+		URL:      fmt.Sprintf("https://%s", client.RancherConfig.Host),
+		TokenKey: token,
+		Insecure: true,
+	}, providerName), nil
+}
+
+// NewSCIMClientWithToken returns a SCIM client configured with the given host, provider, and bearer token
+func NewSCIMClientWithToken(host, providerName, token string) *scimclient.Client {
+	return scimclient.NewClient(&clientbase.ClientOpts{
+		URL:      fmt.Sprintf("https://%s", host),
+		TokenKey: token,
+		Insecure: true,
+	}, providerName)
+}
+
+// WaitForSCIMResourceDeletion polls a SCIM GET endpoint until it returns 404.
+func WaitForSCIMResourceDeletion(getFunc func() (int, error)) error {
+	return kwait.PollUntilContextTimeout(
+		context.Background(),
+		defaults.FiveSecondTimeout,
+		defaults.OneMinuteTimeout,
+		false,
+		func(ctx context.Context) (bool, error) {
+			status, err := getFunc()
+			if err != nil {
+				return false, nil
+			}
+			return status == http.StatusNotFound, nil
+		},
+	)
+}
+
+// CreateSCIMUser creates a SCIM user with a random username and returns the username and ID.
+func CreateSCIMUser(scimClient *scimclient.Client, externalID string, active bool) (string, string, error) {
+	userName := namegen.AppendRandomString("scim-user")
+	user := scimclient.User{
+		Schemas:  []string{scimclient.SCIMSchemaUser},
+		UserName: userName,
+	}
+	if externalID != "" {
+		user.ExternalID = externalID
+	}
+	if active {
+		user.Active = scimclient.BoolPtr(true)
+	}
+	resp, err := scimClient.Users().Create(user)
+	if err != nil {
+		return "", "", err
+	}
+	if resp.StatusCode != http.StatusCreated {
+		return "", "", fmt.Errorf("expected 201, got %d: %s", resp.StatusCode, string(resp.Body))
+	}
+	userID, err := resp.IDFromBody()
+	return userName, userID, err
+}
+
+// CreateSCIMGroup creates a SCIM group with a random displayName and returns the displayName and ID.
+func CreateSCIMGroup(scimClient *scimclient.Client, externalID string) (string, string, error) {
+	groupName := namegen.AppendRandomString("scim-group")
+	group := scimclient.Group{
+		Schemas:     []string{scimclient.SCIMSchemaGroup},
+		DisplayName: groupName,
+	}
+	if externalID != "" {
+		group.ExternalID = externalID
+	}
+	resp, err := scimClient.Groups().Create(group)
+	if err != nil {
+		return "", "", err
+	}
+	if resp.StatusCode != http.StatusCreated {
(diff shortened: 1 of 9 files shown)


actions/scim/scim.go
```diff
@@ -0,0 +1,436 @@
...
+func DisableSCIMFeatureFlag(client *rancher.Client) error {
+	logrus.Info("Disabling SCIM feature flag")
+	return features.UpdateFeatureFlag(client, SCIMFeatureFlag, false)
+}
+
+func IsSCIMFeatureFlagEnabled(client *rancher.Client) (bool, error) {
+	return features.IsEnabled(client, SCIMFeatureFlag)
+}
+
+func ReenableSCIMFeatureFlag(client *rancher.Client, providerName string) error {
```
Priyashetty17: This can be simplified like this: enable the SCIM feature flag if its not already enabled, then call a new helper "WaitForSCIMReady" to handle the polling logic

```
func EnableSCIMAndWait(client *rancher.Client, sc *Client) error {
    enabled, err := features.IsEnabled(client, SCIMFeatureFlag)
    if err != nil {
        return err
    }

    if !enabled {
        if err := features.UpdateFeatureFlag(client, SCIMFeatureFlag, true); err != nil {
            return err
        }
    }

    return WaitForSCIMReady(sc)
}
```

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+	resp1, err := s.scim.Users().Create(user)
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), http.StatusCreated, resp1.StatusCode, "First POST /Users should return 201, body: %s", string(resp1.Body))
+
+	userID, err := resp1.IDFromBody()
+	require.NoError(s.T(), err)
+
+	defer func() {
+		_, _ = s.scim.Users().Delete(userID)
+	}()
```
Priyashetty17: Same comment as other tests: this seems unnecessary, since subSession.Cleanup() already handles cleanup of resources created in the session.

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+	})
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), http.StatusCreated, createResp.StatusCode, "POST /Users should return 201, body: %s", string(createResp.Body))
+
+	userID, err := createResp.IDFromBody()
+	require.NoError(s.T(), err)
+
+	defer func() {
+		_, _ = s.scim.Users().Delete(userID)
+	}()
```
Priyashetty17: This comment applies to ALL the tests doing this: this seems unnecessary, since subSession.Cleanup() already handles cleanup of resources created in the session.

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), http.StatusCreated, createResp.StatusCode, "POST /Groups should return 201, body: %s", string(createResp.Body))
+
+	groupID, err := createResp.IDFromBody()
+	require.NoError(s.T(), err)
+
+	defer func() {
+		logrus.Infof("Deleting SCIM group %s", groupID)
+		_, _ = s.scim.Groups().Delete(groupID)
+	}()
```
Priyashetty17: I have the same comment as in the user deletion tests- this seems unnecessary, since subSession.Cleanup() already handles cleanup of resources created in the session.

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+
+		addResp, err := s.scim.Groups().Patch(groupID, scimclient.PatchOp{
+			Schemas: []string{scimclient.SCIMSchemaPatchOp},
+			Operations: []scimclient.Operation{
+				{Op: "add", Path: "members", Value: []scimclient.Member{{Value: uid}}},
+			},
+		})
+		require.NoError(s.T(), err)
+		require.Equal(s.T(), http.StatusOK, addResp.StatusCode)
+	}
```
Priyashetty17: nit: Instead of using a for loop here, consider creating a helper like `CreateUserAndAddToGroup` in actions/scim, that creates a SCIM user and adds it to a group. This makes the test more readable

```
// CreateUserAndAddToGroup creates a SCIM user and adds it to the given group. Returns the user ID for cleanup.
func CreateUserAndAddToGroup(scimClient *scimclient.Client, groupID, username string) (string, error) {
    userResp, err := scimClient.Users().Create(scimclient.User{
        Schemas:  []string{scimclient.SCIMSchemaUser},
        UserName: username,
    })
    if err != nil {
        return "", err
    }
    if userResp.StatusCode != http.StatusCreated {
        return "", fmt.Errorf("expected 201, got %d: %s", userResp.StatusCode, string(userResp.Body))
    }

    userID, err := userResp.IDFromBody()
    if err != nil {
        return "", err
    }

    addResp, err := scimClient.Groups().Patch(groupID, scimclient.PatchOp{
        Schemas: []string{scimclient.SCIMSchemaPatchOp},
        Operations: []scimclient.Operation{
            {Op: "add", Path: "members", Value: []scimclient.Member{{Value: userID}}},
        },
    })
    if err != nil {
        return "", err
    }
    if addResp.StatusCode != http.StatusOK {
        return "", fmt.Errorf("expected 200, got %d: %s", addResp.StatusCode, string(addResp.Body))
    }

    return userID, nil
}
```

You can then call it in your tests:
```
user1Name := namegen.AppendRandomString("scim-user")
user1ID, err := scimactions.CreateUserAndAddToGroup(s.scim, groupID, user1Name)
require.NoError(s.T(), err)

user2Name := namegen.AppendRandomString("scim-user")
user2ID, err := scimactions.CreateUserAndAddToGroup(s.scim, groupID, user2Name)
require.NoError(s.T(), err)
```

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+	require.NotEmpty(s.T(), crtb.Name)
+
+	foundCRTB, err := rbacapi.GetClusterRoleTemplateBindingsForGroup(s.client, groupPrincipal, s.cluster.ID)
+	require.NoError(s.T(), err, "Should find CRTB for group %s", groupPrincipal)
+	require.Equal(s.T(), groupPrincipal, foundCRTB.GroupPrincipalName)
+	require.Equal(s.T(), s.cluster.ID, foundCRTB.ClusterName)
+	require.Equal(s.T(), "cluster-member", foundCRTB.RoleTemplateName)
+}
+
+func (s *SCIMOpenLDAPTestSuite) TestSCIMUserPaginationMultiPage() {
```
Priyashetty17: This test can be simplified to avoid repeating Users().List and status checks for each page.. Instead, you could create a helper in actions:
```
// ListSCIMUsersPage fetches a SCIM user page and returns the decoded JSON body
func ListSCIMUsersPage(scimClient *scimclient.Client, start, count int) (map[string]interface{}, error) {
    params := url.Values{}
    params.Set("startIndex", fmt.Sprintf("%d", start))
    params.Set("count", fmt.Sprintf("%d", count))

    resp, err := scimClient.Users().List(params)
    if err != nil {
        return nil, err
    }
    if resp.StatusCode != http.StatusOK {
        return nil, fmt.Errorf("expected 200, got %d: %s", resp.StatusCode, string(resp.Body))
    }

    var body map[string]interface{}
    if err := resp.DecodeJSON(&body); err != nil {
        return nil, err
    }
    return body, nil
}
```
Then, in your test, you call
```
b1, err := scimactions.ListSCIMUsersPage(s.scim, 1, 5)
require.NoError(s.T(), err)
```

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+		for _, item := range page {
+			user, _ := item.(map[string]interface{})
+			id, _ := user["id"].(string)
+			require.False(s.T(), seen[id], "user id %s appears on multiple pages", id)
+			seen[id] = true
+		}
+	}
+
+	for _, id := range createdIDs {
+		require.True(s.T(), seen[id], "created user %s should appear in paginated results", id)
```
Priyashetty17: This can be simplified to avoid repeating Users().List and status checks for each page.. Instead, you could create a helper in actions:
```
// ListSCIMUsersPage fetches a SCIM user page and returns the decoded JSON body
func ListSCIMUsersPage(scimClient *scimclient.Client, start, count int) (map[string]interface{}, error) {
    params := url.Values{}
    params.Set("startIndex", fmt.Sprintf("%d", start))
    params.Set("count", fmt.Sprintf("%d", count))

    resp, err := scimClient.Users().List(params)
    if err != nil {
        return nil, err
    }
    if resp.StatusCode != http.StatusOK {
        return nil, fmt.Errorf("expected 200, got %d: %s", resp.StatusCode, string(resp.Body))
    }

    var body map[string]interface{}
    if err := resp.DecodeJSON(&body); err != nil {
        return nil, err
    }
    return body, nil
}
```
Then, in your test, you can call
```
b1, err := scimactions.ListSCIMUsersPage(s.scim, 1, 5)
require.NoError(s.T(), err)
```

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+	respB, err := s.scim.Users().Create(scimclient.User{
+		Schemas:  []string{scimclient.SCIMSchemaUser},
+		UserName: userNameB,
+	})
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), http.StatusCreated, respB.StatusCode)
+
+	userIDB, err := respB.IDFromBody()
+	require.NoError(s.T(), err)
+	defer func() { _, _ = s.scim.Users().Delete(userIDB) }()
```
Priyashetty17: This can be simplified by using the new helper `CreateSCIMUser` suggested in previous tests to create SCIM users.

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+
+	mgmtUserB, err := s.client.Management.User.ByID(userIDB)
+	require.NoError(s.T(), err)
+
+	_, err = rbacapi.VerifyClusterRoleTemplateBindingForUser(s.client, mgmtUserB.ID, 0)
+	require.NoError(s.T(), err, "User B should have no CRTBs")
+}
+
+func (s *SCIMOpenLDAPTestSuite) TestSCIMUsersAsClusterMembers() {
```
Priyashetty17: I have the same comment for this test as the previous tests - use helpers to create user, CRTB

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+	createResp, err := s.scim.Groups().Create(scimclient.Group{
+		Schemas:     []string{scimclient.SCIMSchemaGroup},
+		DisplayName: groupName,
+	})
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), http.StatusCreated, createResp.StatusCode)
+
+	groupID, err := createResp.IDFromBody()
+	require.NoError(s.T(), err)
+	defer func() { _, _ = s.scim.Groups().Delete(groupID) }()
```
Priyashetty17: Same comment as previous tests - Use a helper to create the SCIM group and remove the defer deletion logic from tests

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+	createResp, err := s.scim.Users().Create(scimclient.User{
+		Schemas:  []string{scimclient.SCIMSchemaUser},
+		UserName: userName,
+	})
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), http.StatusCreated, createResp.StatusCode)
+
+	userID, err := createResp.IDFromBody()
+	require.NoError(s.T(), err)
+	defer func() { _, _ = s.scim.Users().Delete(userID) }()
```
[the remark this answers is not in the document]
dasarinaidu: SCIM resources are created via the SCIM HTTP API, not through Rancher's wrangler or management clients. subSession.Cleanup() only tracks resources registered through those clients. The defers are needed to clean up SCIM-provisioned resources that the session is unaware of.

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+			Namespace: prtbNamespace,
+		},
+		ProjectName:      projectName,
+		UserName:         mgmtUser.ID,
+		RoleTemplateName: "project-owner",
+	}
+
+	prtb, err := s.client.WranglerContext.Mgmt.ProjectRoleTemplateBinding().Create(prtbObj)
+	require.NoError(s.T(), err, "Should be able to create PRTB for SCIM user %s", userName)
+	require.NotEmpty(s.T(), prtb.Name)
```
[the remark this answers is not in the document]
dasarinaidu: rbacapi.CreateProjectRoleTemplateBinding and CreateClusterRoleTemplateBinding both call client.AsUser(user). SCIM users don't have a local password so AsUser() always returns 401. I create the bindings directly via WranglerContext.Mgmt to bypass this.

projectapi.CreateProject calls WaitForProjectFinalizerToUpdate internally which timed out (context deadline exceeded) on our test cluster. Reverting to inline WranglerContext.Mgmt.Project().Create() which works reliably.

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+	createResp, err := s.scim.Groups().Create(scimclient.Group{
+		Schemas:     []string{scimclient.SCIMSchemaGroup},
+		DisplayName: groupName,
+	})
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), http.StatusCreated, createResp.StatusCode)
+
+	groupID, err := createResp.IDFromBody()
+	require.NoError(s.T(), err)
+	defer func() { _, _ = s.scim.Groups().Delete(groupID) }()
```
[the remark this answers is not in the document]
dasarinaidu: CreateSCIMUser/CreateSCIMGroup returns only the resource ID. This test needs the resource name in the operation payload itself ("ext-" + userName / userName + "@example.com" / PUT payload). The helper can't be used here without also returning the name, which would change the helper signature. Keeping inline creation so the name is available for the operation.

SCIM resources are created via the HTTP API, not through wrangler/management clients. subSession.Cleanup() does not track them

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
```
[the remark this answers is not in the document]
dasarinaidu: CreateSCIMUser/CreateSCIMGroup returns only the resource ID. This test needs the resource name in the operation payload itself ("ext-" + userName / userName + "@example.com" / PUT payload). The helper can't be used here without also returning the name, which would change the helper signature. Keeping inline creation so the name is available for the operation.

SCIM resources are created via the HTTP API, not through wrangler/management clients. subSession.Cleanup() does not track them

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+
+	createResp, err := s.scim.Groups().Create(scimclient.Group{
+		Schemas:     []string{scimclient.SCIMSchemaGroup},
+		DisplayName: groupName,
+	})
+	require.NoError(s.T(), err)
+	require.Equal(s.T(), http.StatusCreated, createResp.StatusCode)
+
+	groupID, err := createResp.IDFromBody()
+	require.NoError(s.T(), err)
```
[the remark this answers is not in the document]
dasarinaidu: Added CreateSCIMGroup, CreateUserAndAddToGroup, and ListSCIMUsersPage helpers to actions/scim/scim.go as suggested. However, after applying them, none of the tests in this PR can actually use them because:

Tests like TestSCIMUpdateGroup, TestSCIMPatchGroupReplaceExternalID need groupName in the operation payload — the helper returns only the ID
CreateUserAndAddToGroup is not needed yet since the member tests create the user and add them inline
ListSCIMUsersPage wraps logic that's simple enough to stay inline in the pagination test

Removed the three helpers to avoid unused exported functions causing lint failures. They can be added back when a future test actually needs them. CreateSCIMUser and WaitForGroupMemberCount are kept since they are actively used.

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1563 @@
...
+	require.Equal(s.T(), http.StatusCreated, createResp.StatusCode, "POST /Users should return 201, body: %s", string(createResp.Body))
+
+	userID, err := createResp.IDFromBody()
+	require.NoError(s.T(), err, "Created user response should contain an id")
+
+	defer func() {
+		logrus.Infof("Deleting SCIM user %s", userID)
+		_, _ = s.scim.Users().Delete(userID)
+	}()
```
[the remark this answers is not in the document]
dasarinaidu: Added three suite-level helpers:

requireStatus(resp, status, msg) — replaces the 3-line assert+log pattern across all 40 status checks
deleteSCIMUser(id) — replaces all defer func() { _, _ = s.scim.Users().Delete(id) }() patterns
deleteSCIMGroup(id) — same for groups

validation/auth/scim/scim_openldap_test.go
```diff
@@ -0,0 +1,1532 @@
...
+		require.NoError(s.T(), err, "Auth user %s should be able to login", authUser.Username)
+
+		mgmtUser, err := s.client.Management.User.ByID(authAdmin.UserID)
+		require.NoError(s.T(), err, "Should fetch Rancher user for auth user %s", authUser.Username)
+
+		crtbObj := &cattlev3.ClusterRoleTemplateBinding{
+			ObjectMeta:       metav1.ObjectMeta{Namespace: s.cluster.ID, GenerateName: "crtb-"},
+			ClusterName:      s.cluster.ID,
+			UserName:         mgmtUser.ID,
+			RoleTemplateName: "cluster-member",
```
Priyashetty17: nit, consider using existing constant for rbac https://github.com/rancher/tests/blob/main/actions/rbac/rbac.go#L20

Priyashetty17 (dismissed):

Priyashetty17 (approved):

PR #582, opened 2026-03-26
add deprecated pipelines folder to phase out old files

slickwarren:
adding deprecated jenkinsfile folder that we likely want to get rid of in the future, but for now we depend on some of the code/scripts.


rancher-max (dismissed):
This looks fine, but the paths in those Jenkinsfiles look weird since they point to old rancher/rancher paths.

rancher-max:
ughhh gross. We will just have to prioritize getting the non-python rancher-ha-deploy then for teams to transition, because we'll need to strip the Python and old stuff finally. I liked the deprecation idea but don't like pulling all that old stuff over

floatingman (commented):
Probably should ignore the pycache.

floatingman (approved):

lscalabrini01 (approved):

rancher-max:
I don't like just moving these old deprecated things from rancher/rancher to rancher/tests because they have the same issues. However, people needing this can point to your branch directly when the rancher/rancher PR merges so as not to be totally blocked.

slickwarren:
covered in a separate PR

PR #625, opened 2026-04-14
Add HPA validation tests

JonCrowther:
Used Claude Opus to migrate the tests from https://github.com/rancher/rancher/blob/main/tests/validation/tests/v3_api/test_hpa.py to this repo in Go.

There are basic tests:
- Create HPA
- Edit HPA
- Delete HPA

Then for different RBAC roles it checks if the user can get, list, create, delete, update, etc. the horizontal pod autoscaler.


actions/kubeapi/workloads/horizontalpodautoscalers/horizontalpodautoscalers.go
```diff
@@ -0,0 +1,176 @@
...
+
+	return nil
+}
+
+// WaitForHPAActive polls the dynamic client until the HPA reaches active state.
+func WaitForHPAActive(client *rancher.Client, clusterID, namespace, hpaName string) (*autoscalingv2.HorizontalPodAutoscaler, error) {
+	var hpaObj *autoscalingv2.HorizontalPodAutoscaler
+
+	logrus.Infof("Waiting for HPA %s to become active", hpaName)
+	err := kwait.PollUntilContextTimeout(context.TODO(), 5*time.Second, defaults.FiveMinuteTimeout, true, func(ctx context.Context) (done bool, err error) {
```
joesims22: Instead of using time here can you replace this with `defaults.FiveSecondTimeout`.

actions/rbac/hpa.go
```diff
@@ -0,0 +1,223 @@
...
+// DeleteHPA removes an HPA and verifies it no longer exists.
+func DeleteHPA(client *rancher.Client, clusterID, namespaceName, hpaName string) error {
+	logrus.Infof("Deleting HPA %s", hpaName)
+	return hpaapi.DeleteHPA(client, clusterID, namespaceName, hpaName)
+}
+
+// WaitForDeploymentReplicas polls until the deployment has the expected number of available replicas.
+func WaitForDeploymentReplicas(client *rancher.Client, clusterID, namespaceName, deploymentName string, expectedReplicas int32) error {
+	logrus.Infof("Waiting for deployment %s to scale to %d replicas", deploymentName, expectedReplicas)
+	return kwait.PollUntilContextTimeout(context.TODO(), 5*time.Second, defaults.FiveMinuteTimeout, true, func(ctx context.Context) (done bool, err error) {
```
joesims22: Instead of using time here can you replace this with `defaults.FiveSecondTimeout`.

actions/rbac/verify.go
```diff
@@ -242,3 +242,80 @@ func VerifyUserCanRemoveClusterRoles(t *testing.T, client *rancher.Client, user
...
+
+	switch role {
+	case ClusterOwner, ProjectOwner, ProjectMember:
+		assert.Len(t, hpaList.Items, 1)
+		assert.Equal(t, hpa.Name, hpaList.Items[0].Name)
+	case ClusterMember:
+		assert.Empty(t, hpaList.Items)
+	}
+
+	err = DeleteHPA(client, clusterID, namespaceName, hpa.Name)
```
joesims22: Deletion of HPA here can be removed

actions/kubeapi/workloads/horizontalpodautoscalers/create.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/rancher/shepherd/clients/rancher"
+	"github.com/rancher/shepherd/extensions/unstructured"
+	"github.com/rancher/shepherd/pkg/api/scheme"
+	autoscalingv2 "k8s.io/api/autoscaling/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CreateHPA is a helper function that uses the dynamic client to create a horizontal pod autoscaler on a namespace for a specific cluster.
+func CreateHPA(client *rancher.Client, clusterID, namespace string, hpa *autoscalingv2.HorizontalPodAutoscaler) (*autoscalingv2.HorizontalPodAutoscaler, error) {
```
Priyashetty17: We no longer use the dynamic client for the new tests. Instead, we use the Wrangler context for CRUD operations.
Here's an example:  https://github.com/rancher/tests/blob/main/actions/workloads/job/job.go#L20

For this to work for HPA we need to add autoscaler controller to shepherd. I have a PR to add the controller here https://github.com/rancher/shepherd/pull/550

JonCrowther: Why are there 2 different versions of actions https://github.com/rancher/tests/blob/main/actions/kubeapi/workloads/jobs/create.go ?

If we're supposed to use wrangler why keep both?

Priyashetty17: Initially, the tests used the dynamic client to interact with the k8s API, and later the Wrangler context was introduced in Shepherd. For new tests, we have been using the Wrangler context for quite sometime now. Most of the tests use wrangler context now but we still need to clean up some helpers that are using the dynamic client and migrate them to the Wrangler context to avoid confusion.

actions/rbac/verify.go
JonCrowther: done

actions/kubeapi/workloads/horizontalpodautoscalers/create.go
JonCrowther: Thanks for adding the necessary shepherd changes for this. I've updated the PR to use those new clients instead. I also move the action to the right directory to more closely follow the newer format. Let me know if I missed anything :)

actions/rbac/verify.go
```diff
@@ -242,3 +243,80 @@ func VerifyUserCanRemoveClusterRoles(t *testing.T, client *rancher.Client, user
 	err := users.RemoveClusterRoleFromUser(client, user)
 	require.NoError(t, err)
 }
+
+// VerifyUserCanCreateHPA validates that a user with the given role can or cannot create an HPA in the project namespace
+func VerifyUserCanCreateHPA(t *testing.T, client, standardClient *rancher.Client, clusterID, namespaceName string, role Role) {
```
JonCrowther, line 250: So I know a lot of actions have a `verify.go` file, so this might belong to `workloads/horizontalpodautoscaler/verify.go`, but I chose them to be here because really they're verifying the RBAC surrounding HPA, not the HPA object itself. Open to changing, but I personally felt it worked well here

actions/rbac/hpa.go
```diff
@@ -0,0 +1,205 @@
...
+
+	updatedMin := int32(3)
+	updatedMax := int32(10)
+	metrics := []autoscalingv2.MetricSpec{hpaapi.BuildCPUUtilizationMetric(hpaCPUUtilizationValue)}
+	updateObj := hpaapi.NewHPAObject(hpaResp.Name, namespaceName, workload.Name, updatedMin, updatedMax, metrics)
+
+	_, err = hpaapi.UpdateHPA(userClient, clusterID, namespaceName, hpaResp, updateObj)
+	if err == nil {
+		return fmt.Errorf("expected forbidden error when editing HPA, but edit succeeded")
+	}
```
Priyashetty17: Can we explicitly check for forbidden error?
```
_, err = hpaapi.UpdateHPA(userClient, clusterID, namespaceName, hpaResp, updateObj)
if !errors.IsForbidden(err) {
	return fmt.Errorf("expected forbidden error when editing HPA, got: %w", err)
}
```

actions/rbac/hpa.go
```diff
@@ -0,0 +1,205 @@
...
+			corev1.ResourceCPU:    resource.MustParse("100m"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("512Mi"),
+			corev1.ResourceCPU:    resource.MustParse("1000m"),
+		},
+	}
+
+	replicas := 1
+	createdDeployment, err := deploymentapi.CreateDeployment(client, clusterID, namespaceName, replicas, "", "", false, false, false, false)
```
Priyashetty17: CreateHPAWorkload sets resource requests for the container but never applies them to the created deployment, so HPA may not work as expected. You may have to do something like this after deployment is created to use the container resources defined:
```
createdDeployment.Spec.Template.Spec.Containers = []corev1.Container{container}
updatedDeployment, err := deploymentapi.UpdateDeployment(client, clusterID, namespaceName, createdDeployment,true)
if err != nil {
    return nil, fmt.Errorf("failed to update deployment with HPA resources: %w", err)
}
```
Note: UpdateDeployment already waits for the Deployment to become active, so an additional wait call is not required.

actions/workloads/horizontalpodautoscalers/horizontalpodautoscalers.go
```diff
@@ -0,0 +1,238 @@
...
+		if err != nil {
+			return false, err
+		}
+
+		if len(hpaList) == 0 {
+			return true, nil
+		}
+
+		return false, nil
+	})
```
Priyashetty17: Instead of listing, can we use GET here since we are checking the existence of a single HPA?
```
err = kwait.PollUntilContextTimeout(context.Background(),
	defaults.FiveHundredMillisecondTimeout,
	defaults.TenSecondTimeout,
	false,
	func(ctx context.Context) (bool, error) {

		_, err := wranglerContext.Autoscaling.HorizontalPodAutoscaler().Get(namespace, hpaName, metav1.GetOptions{})

		if k8sError.IsNotFound(err) {
			return true, nil
		}

		if err != nil {
			return false, err
		}

		return false, nil
	})
```

actions/workloads/horizontalpodautoscalers/horizontalpodautoscalers.go
```diff
@@ -0,0 +1,238 @@
...
+				Type:         autoscalingv2.AverageValueMetricType,
+				AverageValue: &quantity,
+			},
+		},
+	}
+}
+
+// WatchAndWaitHPAs is a helper function that watches the HPAs
+// sequentially in a specific namespace and waits until they reach a desired conditions state.
+func WatchAndWaitHPAs(client *rancher.Client, clusterID, namespace string, listOptions metav1.ListOptions) error {
```
Priyashetty17: WatchAndWaitHPAs helper function doesn’t appear to be used currently. Should we remove this?

actions/workloads/horizontalpodautoscalers/horizontalpodautoscalers.go
```diff
@@ -0,0 +1,238 @@
...
+// WaitForHPAActive polls the wrangler client until the HPA reaches active state.
+func WaitForHPAActive(client *rancher.Client, clusterID, namespace, hpaName string) (*autoscalingv2.HorizontalPodAutoscaler, error) {
+	var hpaObj *autoscalingv2.HorizontalPodAutoscaler
+
+	logrus.Infof("Waiting for HPA %s to become active", hpaName)
+	err := kwait.PollUntilContextTimeout(context.TODO(), defaults.FiveSecondTimeout, defaults.FiveMinuteTimeout, true, func(ctx context.Context) (done bool, err error) {
+		wranglerContext, err := clusterapi.GetClusterWranglerContext(client, clusterID)
+		if err != nil {
+			return false, nil
+		}
```
Priyashetty17: Can this be moved outside the polling function?

actions/rbac/hpa.go
```diff
@@ -0,0 +1,205 @@
...
+
+	if hpa.Spec.MaxReplicas != expectedMax {
+		return fmt.Errorf("expected maxReplicas %d, got %d", expectedMax, hpa.Spec.MaxReplicas)
+	}
+
+	return nil
+}
+
+// VerifyEditForbidden creates an HPA as the cluster owner, then verifies that the user client cannot edit it.
+func VerifyEditForbidden(userClient, ownerClient *rancher.Client, clusterID, namespaceName string) error {
```
Priyashetty17: Can this be renamed to VerifyUpdateForbidden to better reflect what the function is doing? Also, can this function be moved to rbac/verify.go file?

actions/workloads/horizontalpodautoscalers/horizontalpodautoscalers.go
```diff
@@ -0,0 +1,238 @@
+package horizontalpodautoscalers
```
Priyashetty17: Can we rename the package to hpa for consistency?

actions/rbac/hpa.go
```diff
@@ -0,0 +1,205 @@
...
+	resultHPA, err = hpaapi.WaitForHPAActive(client, clusterID, namespaceName, resultHPA.Name)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return resultHPA, workload, nil
+}
+
+// WaitForDeploymentReplicas polls until the deployment has the expected number of available replicas.
+func WaitForDeploymentReplicas(client *rancher.Client, clusterID, namespaceName, deploymentName string, expectedReplicas int32) error {
```
Priyashetty17: This may not be necessary since there is an existing helper function to wait for deployments to be active that we could use https://github.com/rancher/tests/blob/main/actions/kubeapi/workloads/deployments/deployments.go#L119

actions/rbac/hpa.go
```diff
@@ -0,0 +1,205 @@
...
+		if err != nil {
+			return false, nil
+		}
+
+		return dep.Status.AvailableReplicas >= expectedReplicas, nil
+	})
+}
+
+// VerifyHPAFields validates the fields of an HPA object.
+func VerifyHPAFields(hpa *autoscalingv2.HorizontalPodAutoscaler, expectedName string, expectedMin, expectedMax int32) error {
```
Priyashetty17: Can the verify functions be moved to rbac/verify.go file?

actions/rbac/verify.go
```diff
@@ -242,3 +243,80 @@ func VerifyUserCanRemoveClusterRoles(t *testing.T, client *rancher.Client, user
...
+	require.NoError(t, err)
+
+	hpa, _, err := CreateHPA(standardClient, clusterID, namespaceName, workload)
+	switch role {
+	case ClusterOwner, ProjectOwner, ProjectMember:
+		require.NoError(t, err)
+		err = hpaapi.DeleteHPA(client, clusterID, namespaceName, hpa.Name)
+		assert.NoError(t, err)
+	case ClusterMember:
+		require.Error(t, err)
```
Priyashetty17, line 263: Can we also assert for Forbidden error here? Similar to : https://github.com/JonCrowther/tests/blob/738dc491aba2c70f352a334d0b8f398169f5c1b5/actions/rbac/verify.go#L95-L97

actions/rbac/verify.go
```diff
@@ -242,3 +243,80 @@ func VerifyUserCanRemoveClusterRoles(t *testing.T, client *rancher.Client, user
...
+	hpa, _, err := CreateHPA(client, clusterID, namespaceName, nil)
+	require.NoError(t, err)
+
+	switch role {
+	case ClusterOwner, ProjectOwner, ProjectMember:
+		err = hpaapi.DeleteHPA(standardClient, clusterID, namespaceName, hpa.Name)
+		assert.NoError(t, err)
+	case ClusterMember:
+		err = hpaapi.DeleteHPA(standardClient, clusterID, namespaceName, hpa.Name)
+		assert.Error(t, err)
```
Priyashetty17, line 298: Can we also assert for Forbidden error here?

actions/rbac/verify.go
```diff
@@ -242,3 +243,80 @@ func VerifyUserCanRemoveClusterRoles(t *testing.T, client *rancher.Client, user
...
+
+	hpaList, err := ListHPAsByName(standardClient, clusterID, namespaceName, hpa.Name)
+	require.NoError(t, err)
+
+	switch role {
+	case ClusterOwner, ProjectOwner, ProjectMember:
+		assert.Len(t, hpaList, 1)
+		assert.Equal(t, hpa.Name, hpaList[0].Name)
+	case ClusterMember:
+		assert.Empty(t, hpaList)
```
Priyashetty17, line 320: Can we also assert for Forbidden error here?

actions/rbac/hpa.go
JonCrowther: I've changed it, but I also know we want to prefer `actions/workloads/deployments.go`. Should I make a function in there or just use the existing ones?

actions/workloads/horizontalpodautoscalers/horizontalpodautoscalers.go
JonCrowther: Yep, must have gotten left behind at some point in the moving/re-writes.

actions/workloads/horizontalpodautoscalers/horizontalpodautoscalers.go
JonCrowther: Done, but I simplified to just one `return false, err` since if it's nil we just return nil anyway

actions/rbac/hpa.go
JonCrowther: Looking into this one before I change it

actions/rbac/hpa.go
Priyashetty17: We can reuse the existing one and not create a new helper function. I’ll look into reorganizing some of the kubeapi helper functions in a separate PR.

actions/rbac/verify.go
```diff
@@ -242,3 +245,120 @@ func VerifyUserCanRemoveClusterRoles(t *testing.T, client *rancher.Client, user
...
+
+// VerifyUserCanListHPA validates that a user with the given role can or cannot list HPAs in the project namespace
+func VerifyUserCanListHPA(t *testing.T, client, standardClient *rancher.Client, clusterID, namespaceName string, role Role) {
+	log.Info("Validating if ", role, " can list HPAs in the project namespace.")
+
+	hpa, _, err := CreateHPA(client, clusterID, namespaceName, nil)
+	require.NoError(t, err)
+
+	hpaList, err := ListHPAsByName(standardClient, clusterID, namespaceName, hpa.Name)
+	require.NoError(t, err)
```
Priyashetty17: This assertion runs before the switch, so when a ClusterMember gets a Forbidden error, the test fails early and never reaches the expected IsForbidden check. We should move require.NoError(t, err) inside the switch case for ClusterOwner, ProjectOwner, and ProjectMember.

actions/rbac/verify.go
```diff
@@ -242,3 +245,120 @@ func VerifyUserCanRemoveClusterRoles(t *testing.T, client *rancher.Client, user
...
+// VerifyUserCanDeleteHPA validates that a user with the given role can or cannot delete an HPA in the project namespace
+func VerifyUserCanDeleteHPA(t *testing.T, client, standardClient *rancher.Client, clusterID, namespaceName string, role Role) {
+	log.Info("Validating if ", role, " can delete HPA in the project namespace.")
+
+	hpa, _, err := CreateHPA(client, clusterID, namespaceName, nil)
+	require.NoError(t, err)
+
+	switch role {
+	case ClusterOwner, ProjectOwner, ProjectMember:
+		err = hpaapi.DeleteHPA(standardClient, clusterID, namespaceName, hpa.Name)
```
Priyashetty17: We are calling `hpaapi.DeleteHPA(standardClient, clusterID, namespaceName, hpa.Name)` in each switch case with the same arguments. Can we move the call before the switch and just assert on err per role ?

actions/workloads/deployment/deployment.go
```diff
@@ -122,6 +122,38 @@ func DeleteDeployment(client *rancher.Client, clusterID string, deployment *appv
 	return nil
 }
 
+// UpdateDeployment is a helper to update a deployment using wrangler context
+func UpdateDeployment(client *rancher.Client, clusterID, namespaceName string, deployment *appv1.Deployment, watchDeployment bool) (*appv1.Deployment, error) {
```
Priyashetty17: This can be removed since there is already an existing helper function using the wrangler context that can be reused in the tests: https://github.com/rancher/tests/blob/main/actions/kubeapi/workloads/deployments/update.go#L11 

Note: We plan to reorganize the kubeapi helper functions into kubeapi packages to avoid having to look across multiple packages for related wrangler-based functions, which can cause confusion. As part of this effort, we also plan to replace dynamic client helper functions with wrangler-context-based helpers.

actions/kubeapi/workloads/deployments/update.go
```diff
@@ -14,24 +17,36 @@ func UpdateDeployment(client *rancher.Client, clusterID, namespaceName string, d
...
-	if waitForDeployment {
-		err = WaitForDeploymentActive(client, clusterID, namespaceName, deployment.Name)
+		updatedDeployment, err = wranglerContext.Apps.Deployment().Update(deployment)
 		if err != nil {
-			return nil, err
+			return false, err
 		}
-	}
+
+		if waitForDeployment {
```
Priyashetty17: Does it make sense to move the waitForDeployment outside the retry loop?
Right now we do update -> wait -> retry everything if the wait fails, which means a wait failure can cause the update to be retried even if it already succeeded.

actions/kubeapi/workloads/deployments/update.go
```diff
@@ -14,18 +18,33 @@ func UpdateDeployment(client *rancher.Client, clusterID, namespaceName string, d
 		return nil, err
 	}
 
-	latestDeployment, err := wranglerContext.Apps.Deployment().Get(namespaceName, deployment.Name, metav1.GetOptions{})
-	if err != nil {
-		return nil, err
+	backoff := kwait.Backoff{
```
joesims22, line 20: Nit, could you change this to use the `defaults.FiveSecondTimeout` const

Priyashetty17 (dismissed):
lgtm

actions/kubeapi/workloads/deployments/update.go
JonCrowther: Done

joesims22 (approved):
lgtm

Priyashetty17 (approved):

PR #629, opened 2026-04-17
Update Harvester daily job to pick more tests from daily

lscalabrini01:


slickwarren (dismissed):
if we didn't do this already, can you update the jenkins job to not run in parallel?

lscalabrini01:
> if we didn't do this already, can you update the jenkins job to not run in parallel?

Yes, I can

lscalabrini01:
> if we didn't do this already, can you update the jenkins job to not run in parallel?

It already has been done!

slickwarren (approved):

hamistao (approved):

PR #651, opened 2026-04-28
Add tests for legacy token compatibility to ext tokens

joesims22:
This PR adds test cases to verify the legacy token manager(v3) can accept ext tokens as auth tokens([rancher/rancher #54338](https://github.com/rancher/rancher/issues/54338)).

### DIFF
--- a/actions/tokens/exttokens/ext_tokens.go
+++ b/actions/tokens/exttokens/ext_tokens.go
@@ -191,3 +191,33 @@ func AuthenticateWithExtToken(baseURL, tokenName, tokenValue, apiPath string) er
 	}
 	return nil
 }
+
+// DeleteLegacyTokenWithExtToken sends a raw HTTP DELETE request to the /v3/tokens endpoint using an ext token for Bearer authentication.
+func DeleteLegacyTokenWithExtToken(client *rancher.Client, legacyTokenID string, extTokenValue string) error {
+	deleteURL := fmt.Sprintf("https://%s/v3/tokens/%s", client.RancherConfig.Host, legacyTokenID)
+
+	req, err := http.NewRequest(http.MethodDelete, deleteURL, nil)
+	if err != nil {
+		return fmt.Errorf("failed to create HTTP request: %w", err)
+	}
+
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", extTokenValue))
+
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+	}
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to execute HTTP request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusNoContent {
+		return fmt.Errorf("expected status code 200 or 204, got %d", resp.StatusCode)
+	}
+
+	return nil
+}
--- a/validation/auth/tokens/legacy_token_manager_ext_compatibility_test.go
+++ b/validation/auth/tokens/legacy_token_manager_ext_compatibility_test.go
@@ -1,0 +1,91 @@
+//go:build (validation || infra.any || cluster.any || extended) && !sanity && !stress && !2.8 && !2.9 && !2.10 && !2.11  && !2.12  && !2.13  && !2.14
+
+package tokens
+
+import (
+	"fmt"
+	"strconv"
+	"testing"
+
+	"github.com/rancher/shepherd/clients/rancher"
+	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	"github.com/rancher/shepherd/pkg/session"
+	"github.com/rancher/tests/actions/rbac"
+	"github.com/rancher/tests/actions/settings"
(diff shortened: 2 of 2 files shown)


validation/auth/tokens/ext_token_test.go
```diff
@@ -338,6 +338,29 @@ func (ext *ExtTokenTestSuite) TestDeleteExtTokenAsNonAdmin() {
 	require.True(ext.T(), k8serrors.IsNotFound(err))
 }
 
+func (ext *ExtTokenTestSuite) TestDeleteLegacyTokenUsingExtToken() {
```
Priyashetty17: Joe, you may need to create a new file in the tokens package for these new tests, since they are only applicable for versions 2.15+.
Adding them to the existing test suite would also cause them to run against 2.12, due to the current build tags used in this file

actions/tokens/exttokens/ext_tokens.go
```diff
@@ -191,3 +191,33 @@ func AuthenticateWithExtToken(baseURL, tokenName, tokenValue, apiPath string) er
 	}
 	return nil
 }
+
+// DeleteLegacyTokenWithExtToken sends a raw HTTP DELETE request to the /v3/tokens endpoint using an ext token for Bearer authentication.
+func DeleteLegacyTokenWithExtToken(client *rancher.Client, legacyTokenID string, extTokenValue string) error {
```
Priyashetty17, line 196: Once Das’s PR is merged, do you think this helper could be simplified by leveraging clientbase.Do?
 https://github.com/rancher/shepherd/pull/547/changes#diff-f3bcc8914453bd58079bcee65594b021939b789c4a4d713a0006d65c49a0ce05R20 ? 
something like:
```
resp, err := clientbase.Do(httpClient, "DELETE", deleteURL, nil, map[string]string{
    "Authorization": "Bearer " + extTokenValue,
})
```

validation/auth/tokens/legacy_token_manager_ext_compatibility_test.go
```diff
@@ -0,0 +1,100 @@
...
+	require.NoError(t.T(), err)
+	t.client = client
+
+	log.Info("Getting cluster name from the config file and append cluster details in the struct.")
+	clusterName := client.RancherConfig.ClusterName
+	require.NotEmptyf(t.T(), clusterName, "Cluster name to install should be set")
+	clusterID, err := extensionscluster.GetClusterIDByName(t.client, clusterName)
+	require.NoError(t.T(), err, "Error getting cluster ID")
+	t.cluster, err = t.client.Management.Cluster.ByID(clusterID)
+	require.NoError(t.T(), err)
```
Priyashetty17: Do we need this? It doesn’t appear to be used anywhere.

validation/auth/tokens/legacy_token_manager_ext_compatibility_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	log.Info("Create ext token as standard user")
+	_, standardUserClient, err := rbac.SetupUser(t.client, rbac.StandardUser.String())
+	require.NoError(t.T(), err)
+	standardUserExtToken, err := exttokenapi.CreateExtToken(standardUserClient, t.defaultExtTokenTTL)
+	require.NoError(t.T(), err)
+
+	log.Info("Create an R_SESS cookie containing the ext token value and use it to authenticate a request against the legacy token manager endpoint")
+	standardUserTokenValue := standardUserExtToken.Status.Value
+	baseURL := "https://" + standardUserClient.RancherConfig.Host
```
Priyashetty17: nit, can we instead do: baseURL := fmt.Sprintf("https://%s", standardUserClient.RancherConfig.Host)

validation/auth/tokens/legacy_token_manager_ext_compatibility_test.go
```diff
@@ -0,0 +1,100 @@
...
+	})
+	require.NoError(t.T(), err)
+	log.Infof("Legacy token %s created for standard user %s", standardUserLegacyToken.ID, standardUserClient.UserID)
+
+	log.Info("Create ext token for the standard user")
+	standardUserExtToken, err := exttokenapi.CreateExtToken(standardUserClient, t.defaultExtTokenTTL)
+	require.NoError(t.T(), err)
+
+	log.Info("Using the standard users ext token send a request to delete the standard users legacy token")
+	err = exttokenapi.DeleteLegacyTokenWithExtToken(t.client, standardUserLegacyToken.ID, standardUserExtToken.Status.BearerToken)
```
Priyashetty17: should the token be deleted as a admin or standard user?

joesims22: Good catch, updated to use standard user client

validation/auth/tokens/legacy_token_manager_ext_compatibility_test.go
joesims22: Removed as it is not needed

actions/tokens/exttokens/ext_tokens.go
joesims22: Yes once that is merged I can modify the test to use `clientbase.Do`

Priyashetty17: Since that is the only change, you can create a QA task to revisit it once his PR is merged.

Priyashetty17 (dismissed):
lgtm

Priyashetty17 (approved):

PR #664, opened 2026-05-07
Fix flaky tests and update broken tests

lscalabrini01:


validation/charts/monitoring.go
```diff
@@ -135,46 +136,62 @@ func waitUnknownPrometheusTargets(client *rancher.Client) error {
...
 
-	err := waitUnknownPrometheusTargets(client)
-	if err != nil {
-		return statusInit, err
-	}
+	err := kubewait.ExponentialBackoff(backoff, func() (bool, error) {
+		err := waitUnknownPrometheusTargets(client)
+		if err != nil {
+			logrus.Infof("Waiting for unknown prometheus targets to resolve, retrying: %v", err)
+			return false, nil
```
floatingman: Could you please return the error?

validation/charts/monitoring.go
```diff
@@ -135,46 +136,62 @@ func waitUnknownPrometheusTargets(client *rancher.Client) error {
...
+		}
 
-	bodyString, err := ingresses.GetExternalIngressResponse(client, client.RancherConfig.Host, prometheusTargetsPathAPI, true)
-	if err != nil {
-		return statusInit, err
-	}
+		bodyString, err := ingresses.GetExternalIngressResponse(client, client.RancherConfig.Host, prometheusTargetsPathAPI, true)
+		if err != nil {
+			logrus.Infof("Failed to get prometheus targets response, retrying: %v", err)
+			return false, nil
```
floatingman: Could you please return the error?

validation/charts/monitoring.go
```diff
@@ -135,46 +136,62 @@ func waitUnknownPrometheusTargets(client *rancher.Client) error {
...
+		if err = json.Unmarshal([]byte(bodyString), &mapResponse); err != nil {
+			return false, err
+		}
 
-	if mapResponse["status"] != "success" {
-		return statusInit, errors.New("failed to get targets from prometheus")
-	}
+		if mapResponse["status"] != "success" {
+			logrus.Info("Prometheus targets API did not return success status, retrying")
+			return false, nil
```
floatingman, line 168: Return what the status is.

validation/charts/monitoring.go
```diff
@@ -135,46 +136,62 @@ func waitUnknownPrometheusTargets(client *rancher.Client) error {
...
+		}
 
-	activeTargets := mapResponse["data"].(map[string]interface{})["activeTargets"].([]interface{})
-	if len(activeTargets) < 1 {
-		return false, errors.New("failed to find any active targets")
-	}
+		activeTargets := mapResponse["data"].(map[string]interface{})["activeTargets"].([]interface{})
+		if len(activeTargets) < 1 {
+			logrus.Info("No active prometheus targets found, retrying")
+			return false, nil
```
floatingman, line 174: Return what's in activeTargets.

validation/charts/monitoring.go
floatingman, line 1: Returning the errors for the different checks would help debug future problems with these tests.  Also, add a check to see if Prometheus is responding, and fail faster than retrying for 10 minutes.

validation/longhorn/chartinstall/installation_test.go
```diff
@@ -43,11 +44,11 @@ type LonghornChartTestSuite struct {
...
 }
 
-func (l *LonghornChartTestSuite) TearDownSuite() {
+func (l *LonghornChartTestSuite) TearDownTest() {
 	l.session.Cleanup()
 }
 
-func (l *LonghornChartTestSuite) SetupSuite() {
+func (l *LonghornChartTestSuite) SetupTest() {
 	l.session = session.NewSession()
```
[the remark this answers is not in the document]
floatingman: I think this is a valid point. Does TearDownTest clean up the namespace after each test? If not, then you could solve the issue by generating random namespace names.

validation/longhorn/chartinstall/installation_test.go
```diff
@@ -119,16 +124,6 @@ func (l *LonghornChartTestSuite) TestChartInstall() {
...
-	chart, err := shepherdCharts.GetChartStatus(l.client, l.cluster.ID, charts.LonghornNamespace, charts.LonghornChartName)
-	require.NoError(l.T(), err)
-
-	// If Longhorn was installed by a previous test on this same session, uninstall it to install it again with custom configuration.
-	if chart.IsAlreadyInstalled {
-		l.T().Log("Uninstalling Longhorn as it was installed on a previous test.")
-		err = charts.UninstallLonghornChart(l.client, charts.LonghornNamespace, l.cluster.ID, l.payloadOpts.Host)
-		require.NoError(l.T(), err)
-	}
-
```
floatingman, line 131: Could you explain why you removed the check for an already installed Longhorn chart?

validation/charts/monitoring.go
```diff
@@ -135,46 +136,62 @@ func waitUnknownPrometheusTargets(client *rancher.Client) error {
 // checkPrometheusTargets is a private helper function
 // that checks if all active prometheus targets are healthy by using prometheus API.
 func checkPrometheusTargets(client *rancher.Client) (bool, error) {
+	backoff := kubewait.Backoff{
+		Duration: 1 * time.Minute,
+		Factor:   1.1,
+		Jitter:   0,
+		Steps:    10,
+	}
```
floatingman, line 144: Why such a long retry length? Maybe 30 * time.Second, Factor: 1.5, Jitter: 0.1 would be more appropriate.

floatingman (commented):
Just a few suggestions that will help debug failures in the future.

validation/charts/monitoring.go
hamistao: I think such a high factor would end up taking even more time. Maybe just reducing Duration to 30s is enough to fix this.

validation/charts/monitoring.go
```diff
@@ -135,46 +136,62 @@ func waitUnknownPrometheusTargets(client *rancher.Client) error {
...
+		if mapResponse["status"] != "success" {
+			logrus.Info("Prometheus targets API did not return success status, retrying")
+			return false, nil
+		}
 
-	activeTargets := mapResponse["data"].(map[string]interface{})["activeTargets"].([]interface{})
-	if len(activeTargets) < 1 {
-		return false, errors.New("failed to find any active targets")
-	}
+		activeTargets := mapResponse["data"].(map[string]interface{})["activeTargets"].([]interface{})
```
hamistao: If you have the chance to make the switch, we could switch from `interface{}` to `any` here

validation/charts/monitoring.go
```diff
@@ -135,46 +136,62 @@ func waitUnknownPrometheusTargets(client *rancher.Client) error {
...
 
-	for _, target := range activeTargets {
-		targetMap := target.(map[string]interface{})
-		if targetMap["health"].(string) == "down" {
-			downTargets = append(downTargets, targetMap["labels"].(map[string]interface{})["instance"].(string))
+		var downTargets []string
+		for _, target := range activeTargets {
+			targetMap := target.(map[string]interface{})
+			if targetMap["health"].(string) == "down" {
+				downTargets = append(downTargets, targetMap["labels"].(map[string]interface{})["instance"].(string))
```
hamistao, line 181: Should we be careful as to not get an unexpected failure in case "labels" does not contain a "instance" key? If this is something we can never expect to face then I guess this method is fine.

validation/charts/monitoring.go
```diff
@@ -135,46 +136,62 @@ func waitUnknownPrometheusTargets(client *rancher.Client) error {
...
-	}
-	statusInit = len(downTargets) == 0
 
-	if !statusInit {
-		return statusInit, errors.Wrapf(err, "All active target(s) are not healthy: %v", downTargets)
-	}
+		statusInit = len(downTargets) == 0
+		if !statusInit {
+			logrus.Infof("Prometheus targets not yet healthy, retrying. Down targets: %v", downTargets)
+			return statusInit, nil
```
hamistao: This line can be removed I believe

validation/longhorn/chartinstall/installation_test.go
```diff
@@ -43,11 +44,11 @@ type LonghornChartTestSuite struct {
 	payloadOpts        charts.PayloadOpts
 }
 
-func (l *LonghornChartTestSuite) TearDownSuite() {
+func (l *LonghornChartTestSuite) TearDownTest() {
 	l.session.Cleanup()
 }
 
-func (l *LonghornChartTestSuite) SetupSuite() {
+func (l *LonghornChartTestSuite) SetupTest() {
```
hamistao, line 51: Could you please explain this change?

I see almost all existing suites use `TearDownSuite` and `SetupSuite`, are both ok?

validation/longhorn/chartinstall/installation_test.go
```diff
@@ -78,6 +79,10 @@ func (l *LonghornChartTestSuite) SetupSuite() {
 	latestLonghornVersion, err := l.client.Catalog.GetLatestChartVersion(charts.LonghornChartName, catalog.RancherChartRepo)
 	require.NoError(l.T(), err)
 
+	l.T().Logf("Creating %s namespace", charts.LonghornNamespace)
+	_, err = actionNamespaces.CreateNamespace(client, charts.LonghornNamespace, "{}", map[string]string{}, map[string]string{}, l.project)
+	require.NoError(l.T(), err)
+
```
hamistao, line 85: Amazing that the fix for these tests was just creating the namespace preemptively. Maybe the same logic applies to other chart tests that we have seen failing before.

hamistao (commented):
Most comments are really simple and not that impactful. Thanks for this as we really need to improve some of our tests.

validation/longhorn/chartinstall/installation_test.go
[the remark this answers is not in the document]
lscalabrini01: Yes, the TearDownTest  clean up the namespace after each test

validation/charts/monitoring.go
lscalabrini01: done

validation/charts/monitoring.go
lscalabrini01: done

validation/charts/monitoring.go
lscalabrini01: I can log the status

validation/charts/monitoring.go
lscalabrini01: done

validation/charts/monitoring.go
lscalabrini01: I can log the data

validation/charts/monitoring.go
lscalabrini01: I guess it dosen't contain a "instance" key

validation/charts/monitoring.go
lscalabrini01: done

validation/longhorn/chartinstall/installation_test.go
lscalabrini01: The SetupTest and TearDownTest run the setup and cleanup for each test, the SetupSuite and TearDownSuite execute once by Test Suite, for this test I need to setup and cleanup the resources for each test

validation/charts/monitoring.go
lscalabrini01: done

validation/longhorn/chartinstall/installation_test.go
lscalabrini01: It’s because it will be uninstalled by the cleanup anyway

slickwarren (approved):

floatingman (approved):

validation/longhorn/chartinstall/installation_test.go
hamistao: Oh got it, thanks for elaborating

hamistao (approved):

PR #665, opened 2026-05-08
Test Result improvements in Qase

lscalabrini01:
This PR improves Qase reporting metadata for validation pipelines by capturing the rancher/tests git commit SHA in Jenkins and including it in the Qase test run description, along with a small adjustment to captured image output formatting.

Changes:

- Add RANCHER_TEST_COMMIT_ID env var support and include it in the Qase run description (reporter v2).
- Populate RANCHER_TEST_COMMIT_ID in the recurring Jenkins pipeline using git rev-parse HEAD.
- Adjust how pulled image names are written by the image-capture utility.


scripts/capture_images/capture_images.go
```diff
@@ -89,7 +89,7 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID
 			matches := pulledRegex.FindStringSubmatch(k8sEvent.Message)
 
 			if len(matches) > 1 {
-				file.WriteString(fmt.Sprintf("%s (pulled during test)\n", matches[1]))
+				file.WriteString(fmt.Sprintf("%s \n", matches[1]))
```
hamistao: I would like to ask you to revert this change only since I am making a major refactoring of this script and I would prefer to include this kind of change there.
Plus, having a simple way to distinguish images that were already present on the cluster from the ones that were pulled during the test is something nice to have.

hamistao (commented):
Looks great, aside from one comment

scripts/capture_images/capture_images.go
lscalabrini01: Yes, but we have a limit of characters on the QASE description, one thing that we could do is keep distinguish images on the job logs but just same the images versions on the QASE description

hamistao: If the issue is character limits, I believe this will be addressed along with the changes I am working on since they should make the overall output shorter. Also, we are bound to face this issue anyway if we ever get a test run that use a but more images or that uses a cluster has more preexisting stuff on it. So TLDR, I think we should chase another way to address this issue. We can talk more about this synchronously today.

hamistao (dismissed):

hamistao (approved):

floatingman (approved):

PR #674, opened 2026-05-12
Improve image capture logic

hamistao:
This makes a few changes to the capture images logic, both on the go script and the Jenkinsfile. These changes amount to:
 - Ensure we report all the images;
 - Simplfies the process overall by writing a full report on a single file instead of having a separate file for each cluster. As a part of this simplification the order of operations had to be changed from: (1- run tests, 2- qase report, 3- stop imageCapturer and show images) to (1- run tests, 2- stop imageCapturer and show images, 3-qase report). This way we can be sure we all the images have been recorded on the file before creating the qase report.
 - Changes the format of the images output, more info on the specific changes on private channels;
 - Show kubernetes version and previous existing images for all the clusters and use pure shepherd instead of kubectl to get this information;

### DIFF
--- a/scripts/capture_images/capture_images.go
+++ b/scripts/capture_images/capture_images.go
@@ -7,16 +7,18 @@ import (
 	"os"
 	"os/signal"
 	"regexp"
-	"strings"
+	"slices"
+	"sort"
 	"sync"
 	"syscall"
 
 	"github.com/rancher/shepherd/clients/rancher"
-	"github.com/rancher/shepherd/extensions/clusters"
+	clusterV3 "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
+	"github.com/rancher/shepherd/extensions/defaults/stevetypes"
 	"github.com/rancher/shepherd/extensions/kubeconfig"
-	"github.com/rancher/shepherd/extensions/kubectl"
 	"github.com/rancher/shepherd/extensions/rancherversion"
 	"github.com/rancher/shepherd/pkg/session"
+	"github.com/rancher/tests/actions/workloads/pods"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -24,10 +26,8 @@ import (
 )
 
 const (
-	logBufferSize            = "2MB"
-	kubernetesVersionCommand = "kubectl version -o json | jq -r '\"kubernetes:\" + .serverVersion.gitVersion'"
-	imagesVersionsCommand    = "kubectl get pods --all-namespaces -o jsonpath=\"{..image}\" |tr -s '[[:space:]]' '\n' |sort |uniq"
-	imagesPath               = "/app/images/"
+	localClusterName = "local"
+	imageReportPath  = "/app/image-report/image-report.txt"
 )
 
 var (
@@ -35,14 +35,26 @@ var (
 	existingRegex = regexp.MustCompile(`Container image "([^"]+)" already present on machine`)
 )
 
-type ClusterInfo struct {
-	Name string
-	ID   string
+// writeImageSet writes a set of image names (in the form of keys on a map) to a file in sorted fashion.
+func writeImageSet(imageSet map[string]string, file *os.File) {
+	var sortedImages []string
+
+	for image, value := range imageSet {
+		sortedImages = append(sortedImages, image+value)
+	}
+
+	sort.Slice(sortedImages, func(i, j int) bool {
+		return sortedImages[i] < sortedImages[j]
+	})
+
+	for _, image := range sortedImages {
+		file.WriteString(image + "\n")
+	}
 }
 
 // Connects to the specified Rancher managed Kubernetes cluster, monitoring and parsing its events while the test
-// is run. Writes all pulled image names to a file.
-func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID string, file *os.File) error {
+// is run. Stores all unique used image names within a provided map.
+func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID string, imageSet map[string]string) error {
 	clientConfig, err := kubeconfig.GetKubeconfig(client, clusterID)
 	if err != nil {
 		return fmt.Errorf("Failed building client config from string: %v", err)
@@ -89,14 +101,18 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID 
 			matches := pulledRegex.FindStringSubmatch(k8sEvent.Message)
 
 			if len(matches) > 1 {
-				file.WriteString(fmt.Sprintf("%s (pulled during test)\n", matches[1]))
+				imageSet[matches[1]] = "(pulled during test)"
 				continue
 			}
 
 			matches = existingRegex.FindStringSubmatch(k8sEvent.Message)
 
-			if len(matches) > 1 {
-				file.WriteString(fmt.Sprintf("%s \n", matches[1]))
+			if len(matches) <= 1 {
+				continue
+			}
+
+			if _, ok := imageSet[matches[1]]; !ok {
+				imageSet[matches[1]] = ""
 			}
 		}
 	}
@@ -104,7 +120,13 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID 
 
 func main() {
 	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, syscall.SIGTERM)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	reportFile, err := os.Create(imageReportPath)
+	if err != nil {
+		panic(fmt.Errorf("Error creating report file: %w", err))
+	}
+	defer reportFile.Close()
 
 	client, err := rancher.NewClient("", session.NewSession())
 	if err != nil {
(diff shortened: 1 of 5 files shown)


lscalabrini01 (dismissed):

scripts/capture_images/capture_images.go
```diff
@@ -7,42 +7,34 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
 const (
-	logBufferSize            = "2MB"
-	kubernetesVersionCommand = "kubectl version -o json | jq -r '\"kubernetes:\" + .serverVersion.gitVersion'"
-	imagesVersionsCommand    = "kubectl get pods --all-namespaces -o jsonpath=\"{..image}\" |tr -s '[[:space:]]' '\n' |sort |uniq"
-	imagesPath               = "/app/images/"
```
lscalabrini01, line 30: Could we keep saving the logs on the file for the report? the report is using these file to read the logs and save on qase description

lscalabrini01 (changes requested):

lscalabrini01 (dismissed):

scripts/capture_images/capture_images.go
hamistao: Fixed by changing the approach to writing to a single file.

lscalabrini01 (dismissed):

slickwarren (dismissed):

scripts/capture_images/capture_images.go
```diff
@@ -192,32 +189,13 @@ func main() {
...
-	var b strings.Builder
-
-	config, err := rancherversion.RequestRancherVersion(client.RancherConfig.Host)
-	if err != nil {
-		return "", err
-	}
 
-	b.WriteString(fmt.Sprintf("rancher:%s\n", config.RancherVersion))
-	b.WriteString(fmt.Sprintf("rancher-commit:%s\n", config.GitCommit))
-	b.WriteString(fmt.Sprintf("is-prime:%t\n", config.IsPrime))
```
floatingman, line 210: Just curious. Why did you stop gathering all the Rancher version information?

scripts/capture_images/capture_images.go
```diff
@@ -89,97 +82,101 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID
...
+		if err != nil {
+			panic(fmt.Errorf("Failed getting pods for cluster %s: %w", cluster.Name, err))
 		}
-	}
 
-	c, err := clusters.NewClusterMeta(client, clusterName)
-	if err != nil {
-		panic(fmt.Errorf("Failed to create file for versions: %v", err))
-	}
+		imageSet := make(map[string]struct{})
```
floatingman: Why is this variable named the same as the one on line 170? I know this one is scoped to just this block, but it may be confusing to see the same variable being created and assigned in two different places.

scripts/capture_images/capture_images.go
```diff
@@ -89,97 +82,101 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID
...
 		channelList = append(channelList, doneChan)
 
+		imageSet[clusterInfo.Name] = make(map[string]string)
+
 		go func() {
-			file = filesMap[clusterInfo.Name]
-			err := connectAndMonitor(client, doneChan, clusterInfo.ID, file)
+			err := connectAndMonitor(client, doneChan, clusterInfo.ID, imageSet[clusterInfo.Name])
 			if err != nil {
 				panic(fmt.Errorf("Failed to capture used images: %v", err))
```
floatingman, line 205: Could you explain why this panic is left, while the others were changed to log the error? This will cause the whole process to crash if one cluster connection fails.

scripts/capture_images/capture_images.go
```diff
@@ -89,97 +82,101 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID
...
 }
 
 func main() {
 	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, syscall.SIGTERM)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	reportFile, err := os.Create(imageReportPath)
+	if err != nil {
+		panic(fmt.Errorf("Error creating report file: %w", err))
```
floatingman, line 127: Could you explain why this panic is left, while the others were changed to log the error? This will cause the whole process to crash if the report fails to write.

scripts/capture_images/capture_images.go
```diff
@@ -89,97 +82,101 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID
...
-	if clusterName != "" {
-		localClusterID, err := clusters.GetClusterIDByName(client, "local")
-		if err != nil {
-			panic(fmt.Errorf("Error getting local cluster ID: %w", err))
-		}
 
-		clusterList = append(clusterList, ClusterInfo{Name: "local", ID: localClusterID})
+	allClusters, err := client.Management.Cluster.List(nil)
+	if err != nil {
+		panic(fmt.Errorf("Failed getting cluster objects: %w", err))
```
floatingman, line 140: Could you explain why this panic is left, while the others were changed to log the error? This will cause the entire process to crash if a single cluster object fails.

scripts/capture_images/capture_images.go
```diff
@@ -89,97 +82,101 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID
...
+		targetClusters = allClusters.Data
+	}
+
+	for _, cluster := range targetClusters {
+		reportFile.WriteString("Kubernetes version on " + cluster.Name + " cluster: " + cluster.Version.GitVersion + "\n")
+
+		downstreamClient, err := client.Steve.ProxyDownstream(cluster.ID)
 		if err != nil {
-			panic(fmt.Errorf("Failed getting local cluster ID: %w", err))
+			panic(fmt.Errorf("Failed creating downstream client for cluster %s: %w", cluster.Name, err))
```
floatingman, line 168: Could you explain why this panic is left, while the others were changed to log the error? This will cause the whole process to crash if one downstream client fails to create.

scripts/capture_images/capture_images.go
```diff
@@ -89,97 +82,101 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID
...
 		}
 
-		for _, cluster := range allClusters.Data {
-			clusterList = append(clusterList, ClusterInfo{cluster.Name, cluster.ID})
+		steveClient := downstreamClient.SteveType(stevetypes.Pod)
+		clusterPods, err := steveClient.List(nil)
+		if err != nil {
+			panic(fmt.Errorf("Failed getting pods for cluster %s: %w", cluster.Name, err))
```
floatingman, line 174: Could you explain why this panic is left, while the others were changed to log the error? This will cause the whole process to crash if one pod fails to retrieve.

validation/Jenkinsfile
```diff
@@ -130,53 +129,52 @@ node {
...
+                  -v ./validation/image-report:/app/image-report -v ./results:/app/results ${imageName} \
+                  sh -c "${rootPath}/pipeline/scripts/${reporterScript}; if [ -f ./reporter ]; then ./reporter; else echo 'Reporter script not present'; fi"
+                """
               }
               stage('Test Report') {
                 sh "docker cp ${testContainer}:${rootPath}${testResultsOut} ."
                 step([$class: 'JUnitResultArchiver', testResults: "**/${testResultsOut}"])
                 sh "docker stop ${testContainer}"
                 sh "docker rm -v ${testContainer}"
+                sh "docker rm -v ${reporterContainer}"
```
floatingman, line 181: Why aren't you removing this container in the outer catch block? You can create a try/catch block in this stage to solve this.

validation/Jenkinsfile
```diff
@@ -130,53 +129,52 @@ node {
...
                   }
-                } catch(err) {
-                  echo 'Test run had failures. Collecting results...'
                 }
+
+                sh """
+                  docker run --name ${reporterContainer} -t --env-file ${envFile} \
+                  -v ./validation/image-report:/app/image-report -v ./results:/app/results ${imageName} \
+                  sh -c "${rootPath}/pipeline/scripts/${reporterScript}; if [ -f ./reporter ]; then ./reporter; else echo 'Reporter script not present'; fi"
+                """
```
floatingman: This needs to be in a try/catch block, because if it fails, the whole build fails at this point, never making it to the Test Report stage.

scripts/capture_images/capture_images.go
```diff
@@ -89,97 +82,101 @@ func connectAndMonitor(client *rancher.Client, sigChan chan struct{}, clusterID
...
-	file.Write([]byte(versions + "\n"))
+		reportFile.WriteString("Images present on the " + cluster.Name + " cluster previous to the test run:\n\n")
 
-	filesMap := make(map[string]*os.File)
-	for _, clusterInfo := range clusterList {
-		file, err := os.Create(imagesPath + clusterInfo.Name)
-		if err != nil {
-			panic(fmt.Errorf("Failed to create file for image names: %v", err))
+		for image := range imageSet {
+			reportFile.WriteString(image + "\n")
```
floatingman: Just so you know, maps in Go are always in random order, so it's a good idea to sort the image set before writing the image, so the report is always in the same order.

scripts/capture_images/capture_images.go
```diff
@@ -192,32 +189,13 @@ func main() {
...
-	b.WriteString(fmt.Sprintf("is-prime:%t\n", config.IsPrime))
-
-	versionsCommand := []string{
-		"sh", "-c", fmt.Sprintf("%s && %s", kubernetesVersionCommand, imagesVersionsCommand),
-	}
-
-	log, err := kubectl.Command(client, nil, clusterID, versionsCommand, logBufferSize)
-	if err != nil {
-		return "", err
+	for clusterName, images := range imageSet {
```
floatingman, line 218: Same thing about the random order. It's a good idea to sort before this for loop.

validation/pipeline/qase/reporter-v2/main.go
```diff
@@ -30,15 +29,15 @@ var (
...
 )
 
 const (
 	// Doc: https://developers.qase.io/reference/get-cases
 	requestLimit = 100
 	// Doc: https://developers.qase.io/reference/create-run
 	descriptionLimit = 10000
-	imagesPath       = "/app/images/"
+	imageReportPath  = "/app/image-report/image-report.txt"
+	testResultsJSON  = "/app/results/results.json"
```
floatingman, line 40: There used to be a constant setup with this path.

scripts/capture_images/capture_images.go
```diff
@@ -192,32 +189,13 @@ func main() {
...
-		return "", err
-	}
 
-	b.WriteString(fmt.Sprintf("rancher:%s\n", config.RancherVersion))
-	b.WriteString(fmt.Sprintf("rancher-commit:%s\n", config.GitCommit))
-	b.WriteString(fmt.Sprintf("is-prime:%t\n", config.IsPrime))
-
-	versionsCommand := []string{
-		"sh", "-c", fmt.Sprintf("%s && %s", kubernetesVersionCommand, imagesVersionsCommand),
-	}
```
lscalabrini01, line 214: Same question, I guess we should keep collecting the rancher and kubernetes versions and is_prime and commit id informations

scripts/capture_images/capture_images.go
hamistao: Good idea, for big lists of images this should make it easier to read. I will do the same for the list of images used before the tests.

validation/pipeline/qase/reporter-v2/main.go
hamistao: Sorry I don't think I folllow this one.

validation/Jenkinsfile
hamistao: Good catch, will fix.

validation/Jenkinsfile
hamistao: Good point, I will fix this

scripts/capture_images/capture_images.go
hamistao: I can rename it to clusterImageSet

scripts/capture_images/capture_images.go
hamistao: I will resolve the other comments just like this one and just focus the discussion here, even though this specific panic definetely makes sense here as if this fails nothing that comes after can do their job. But I do see your point with some of the others.

I don't know which panics you are referring to that have been changed. Nonetheless, this behavior makes sense to me as an incomplete report is not necessarily better than no report.

scripts/capture_images/capture_images.go
hamistao: This was a mistake on my part, this was not meant to be deleted.

@lscalabrini01

validation/pipeline/qase/reporter-v2/main.go
floatingman: Don't worry about it. There is a constant called TestResultsJSON in the qase actions that is just "results.json" so I wouldn't worry about changing this.
[defaults.go](https://github.com/rancher/tests/blob/a95ab294cfcc5a2d4473a360d7ccd07e56af200a/actions/qase/defaults.go#L19)

hamistao: Oh got it now, thanks!

scripts/capture_images/capture_images.go
hamistao: Done

floatingman (approved):

lscalabrini01 (approved):

PR #691, opened 2026-05-21
Improve capture images logic

hamistao:
The previous attempt at this (https://github.com/rancher/tests/pull/674) was reverted due to problems caused as a result of changing the results.json path in the Qase reporter (https://github.com/rancher/tests/pull/688). This time aroung, a similar method was chosen but this time the results path was preserved.

Besides this, this PR also fixes identation on  `validationJenkinsfile` and removes one useless try/catch block.

### DIFF
--- a/validation/pipeline/qase/reporter-v2/main.go
+++ b/validation/pipeline/qase/reporter-v2/main.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"slices"
 	"strconv"
 	"strings"
 
@@ -38,7 +37,8 @@ const (
 	requestLimit = 100
 	// Doc: https://developers.qase.io/reference/create-run
 	descriptionLimit = 10000
-	imagesPath       = "/app/images/"
+	imageReportPath  = "/app/image-report/image-report.txt"
+	testResultsJSON  = "results.json"
 )
 
 func main() {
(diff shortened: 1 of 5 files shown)


validation/pipeline/qase/reporter-v2/main.go
```diff
@@ -38,7 +37,8 @@ const (
 	requestLimit = 100
 	// Doc: https://developers.qase.io/reference/create-run
 	descriptionLimit = 10000
-	imagesPath       = "/app/images/"
+	imageReportPath  = "/app/image-report/image-report.txt"
+	testResultsJSON  = "results.json"
```
floatingman, line 41: So here is where I commented last time about the constant that is set up with this. You actually used the constant this in reporter/main.go

floatingman (approved):

validation/pipeline/qase/reporter-v2/main.go
hamistao: Oh this is what you meant :laughing:

hamistao: So we are all good this time

floatingman: I mean, you can add the constant here, like in reporter/main.go for consistency's sake.

lscalabrini01 (approved):

PR #693, opened 2026-05-21
Elemental Tests

lscalabrini01:
This PR updates the test tagging and supporting metadata to introduce/standardize the pit.elemental build tag (replacing the prior pit.elemental.daily usage), extends Elemental coverage to additional validation suites, and wires Elemental-related tagging/schema information into supporting pipeline and Qase schema files. It also updates machine pool role matching to better support combined-role pools during scaling.

Changes:

- Standardize Go build tags across many validation tests to use pit.elemental and update tag documentation accordingly.
- Add/adjust PIT schema mappings (including a new NeuVector hardened schema entry) and propagate validation tag parameters through the recurring multibranch Jenkins pipeline.
- Update machine pool role matching logic to support subset matches and add an explicit error when no matching pool is found.


validation/neuvector/neuvector_hardened_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build (validation || infra.any || i.cluster.any || sanity || pit.daily || pit.elemental.daily) && !stress && !extended
+//go:build (validation || infra.any || i.cluster.any || sanity || pit.event) && !stress && !extended
```
floatingman, line 1: is pit.event correct?

floatingman (approved):

validation/neuvector/neuvector_hardened_test.go
navsalehi: I had the same question. Most of the changes in this PR replace pit.elemental.daily with pit.elemental, but this test was switched to pit.event. Was this intentional?

lscalabrini01: Yes, it is correct https://github.com/rancher/tests/blob/main/TAG_GUIDE.md?plain=1#L110

slickwarren (approved):

navsalehi (approved):

validation/neuvector/neuvector_hardened_test.go
lscalabrini01: Yes, it was intentional

PR #696, opened 2026-05-26
update shepherd and pkg/api versions

Priyashetty17:
**Version bumps:**
- rancher/rancher/pkg/apis
- rancher/shepherd
- rancher/rancher
- sigs.k8s.io/cluster-api: v1.8.3 → v1.12.2 to match what's in rancher/rancher

**Code changes:**
- ranchermonitoring.go: Removed deprecated RKE1 support (the KubernetesProviderRKE conditional that was causing a compile error)
- verify.go: Updated cluster-api import path from api/v1beta1 → api/core/v1beta2 (to match cluster-api v1.12.2)


joesims22 (dismissed):

Priyashetty17:
@rancher/qa-pit-crew There is a failure in https://github.com/rancher/tests/blob/main/actions/ssh/ssh.go after updating the pkg/apis version since this file still references RKE1 fields . The newer pkg/apis version in rancher/rancher removed NodeConfig from NodeStatus (as part of the RKE1 field cleanup), so the struct no longer includes NodeConfig.. CreateSSHNode function is used only in tests/validation/fleet/upgrade/upgrade.go . Is that a PIT test? If yes, could you fix the test to remove RKE1 references?

markusewalker (dismissed):

actions/provisioning/verify.go
```diff
@@ -703,12 +703,9 @@ func VerifyHostnameLength(t *testing.T, client *rancher.Client, clusterObject *s
 		dynamic, err := client.GetRancherDynamicClient()
 		require.NoError(t, err)
 
-		gv, err := schema.ParseGroupVersion(capiMachine.Spec.InfrastructureRef.APIVersion)
-		require.NoError(t, err)
-
 		gvr := schema.GroupVersionResource{
-			Group:    gv.Group,
-			Version:  gv.Version,
+			Group:    capiMachine.Spec.InfrastructureRef.APIGroup,
```
markusewalker: Talked offline of this, but this section is breaking the hostname truncation test, doing some testing offline to get to the root cause.

hamistao: Hey! Can you tell if this is related to this failure on CI: https://github.com/rancher/tests/actions/runs/26466518447/job/77928171200?pr=696?

hamistao: Looks like the CI failure is actually related to the go.mod update and some change on `github.com/rancher/rancher/pkg/apis/management.cattle.io/v3`. Looking more into this

Priyashetty17: This is a different issue and not related to the failure with the ssh helper

actions/go.mod
```diff
@@ -8,8 +8,8 @@ replace (
 	github.com/docker/distribution => github.com/docker/distribution v2.8.2+incompatible // rancher-machine requires a replace is set
 	github.com/docker/docker => github.com/docker/docker v20.10.27+incompatible // rancher-machine requires a replace is set
 
-	github.com/rancher/rancher/pkg/apis => github.com/rancher/rancher/pkg/apis v0.0.0-20260105201356-c4811cb9f2af
-	github.com/rancher/rancher/pkg/client => github.com/rancher/rancher/pkg/client v0.0.0-20260105201356-c4811cb9f2af
+	github.com/rancher/rancher/pkg/apis => github.com/rancher/rancher/pkg/apis v0.0.0-20260526130302-36fd88c8ebfc
```
hamistao, line 11: I believe for this change to not break the ssh helper we can simply use `v3Node.Spec.CustomConfig.User` instead of `v3Node.Status.NodeConfig.User` on `actions/ssh/ssh.go`. Let's see after fixing this is more problems will arise.

hamistao (commented):
Just one comment regarding the CI failure, aside from that LGTM.

actions/go.mod
hamistao: @Priyashetty17 If you want I can contribute to this PR directly and try to fix this.

Priyashetty17: @hamistao  Thanks for looking into this. The update itself isn’t breaking the helper. The upstream API change in rancher/rancher was intentional and introduced as part of this commit: https://github.com/rancher/rancher/commit/1f22fcaec55b38791b23937ba5a7b8d6a545d380 

We now need to update the helpers to remove the RKE1 references. Since the SSH helper is currently only used by the PIT team, could you please help update it to remove the RKE1-specific references so CI can pass again?

actions/provisioning/verify.go
Priyashetty17: @markusewalker To avoid issues with the versions, I replaced the dynamic client with Steve API. Could you please check whether this resolves the issue you are seeing?

markusewalker (approved):

actions/provisioning/verify.go
markusewalker: Resolved, thanks Priya! The specific test is working again, resolving comment.

hamistao (approved):
Approved by the PIT Team!

actions/go.mod
Priyashetty17: Resolving the comment since this issue has been addressed

PR #712, opened 2026-06-05
Replace dynamic client with wrangler based helpers and extensions

Priyashetty17:
This PR replaces dynamic client helpers with wrangler-based helpers and new extensions (introduced recently in shepherd) for some of the functionalities. 

- RBAC
- Secrets
- ConfigMaps
- Ext Tokens
- Ext Kubeconfigs 
- HPA
- Public API Users
- Workloads

`Note:`
- Having a mix of dynamic client and Wrangler based helpers has caused confusion, especially for new test writers. This PR is part of a broader effort https://github.com/rancher/qa-tasks/issues/2323 to refactor and organize the kubeapi helpers to reduce that confusion.
- The impacted tests are mostly Collie owned test suites. However, all the impacted tests will be run to ensure that none of the tests are broken. 
- Key Changes: Removed dynamic client based helpers in `actions/workloads/`, `actions/kubeconfigs/`, `actions/users/`, `actions/settings/` and consolidated them into `actions/kubeapi/` hierarchy


actions/kubeapi/namespaces/verify.go
```diff
@@ -161,60 +161,78 @@ func VerifyNamespaceResourceQuotaValidationStatus(client *rancher.Client, cluste
...
+			return false, nil
+		}
 
-	return nil
+		if int(actualPodLimit.Value()) != expectedPodLimit {
+			return false, nil
+		}
+
+		return true, nil
+	})
```
hamistao: could be `return int(actualPodLimit.Value()) == expectedPodLimit, nil` perhaps, do you like this?

actions/kubeapi/namespaces/verify.go
```diff
@@ -161,60 +161,78 @@ func VerifyNamespaceResourceQuotaValidationStatus(client *rancher.Client, cluste
...
 }
 
 // VerifyNamespacePodQuotaValidationStatus checks if the resource quota annotation in a namespace matches the expected pod limits and validation status.
```
hamistao, line 185: I liked the `..., with polling.` addition in the previous function

actions/kubeapi/kubeconfigs/ext_kubeconfigs.go
```diff
@@ -0,0 +1,161 @@
+package kubeconfigs
```
hamistao, line 1: I like the standards set by these files, keeping data structure handling/error handling/polling in the actions package and the proper interaction with Rancher in shepherd. TA!

actions/projects/projects.go
```diff
@@ -12,7 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// GetProjectByName is a helper function that returns the project by name in a specific cluster.
+// GetProjectByName is a helper function that returns the project (norman) by name in a specific cluster.
```
hamistao, line 15: Just make sure I understand the scope of this effort, do we want wrangler over norman here and thus use `GetProjectByName` from `shepherd/extensions/kubeapi/projects` instead of this function at some point?

hamistao (commented):
Thanks a lot for this. Just leaving a couple of simple suggestions and questions.

actions/projects/projects.go
Priyashetty17: With this PR, the actions/kubeapi package helpers use the k8s public API, while code outside that package uses Steve or Norman. The public API for projects was introduced in 2.8 but the UI still uses Norman. If your tests are intended to align with the UI behavior, use Norman otherwise use the public API shepherd/extensions/kubeapi/projects .
This helper uses Norman. I didn't modify any Steve/Norman helpers as part of this PR, I only updated the comment to explicitly state that this function uses Norman rather than Steve.

actions/kubeapi/namespaces/verify.go
Priyashetty17: Thanks , updated

hamistao (approved):
TA!

joesims22 (approved):
lgtm

PR #714, opened 2026-06-10
Bump k8s and shepherd versions

Priyashetty17:
- Bump shepherd version to latest
- Bump k8s version to match rancher/rancher
- Update Dockerfiles to golang:1.26: Bumping shepherd version also bumped Go to 1.26.0 (shepherd's [go.mod](https://github.com/rancher/shepherd/blob/main/go.mod)). The Jenkins build images were still on golang:1.25, which would break, so this PR bumps the build images to 1.26
- Bumped golangci-lint to v2.12.1 (built with Go 1.26 and matches what's in rancher/shepherd) so CI lint works with the new go 1.26 modules; the previous v2.7.2 was built with Go 1.25 and failed.


joesims22 (approved):

hamistao (approved):
First of all, thanks a lot for this, we were in need of some updates.

Of course, if test happens to break because of a deprecated function we should just update it. But have you tried running something using the new versions to scout for something like this?

@Priyashetty17

lscalabrini01 (approved):

Priyashetty17:
@hamistao Good question. I ran a validation test against these changes to check for any version compatibility issues  and it completed successfully without any issues. I will share the freeform job with you offline.

PR #722, opened 2026-06-15
Fixing harvester test cleanup

lscalabrini01:
This PR adjusts the Longhorn validation suite teardown and strengthens a few assertions in the StatefulSet/PVC scaling test to reduce cleanup side effects and improve failure clarity.

Changes:

- Removed explicit Longhorn chart uninstall logic from TearDownSuite, relying on session cleanup instead.
- Added require.NotEmpty checks after listing nodes, PVCs, and volumes to fail earlier with clearer errors.
- Renamed pod list variables for clarity (pods → podList, newPods → newPodList).

### DIFF
--- a/validation/longhorn/longhorn_test.go
+++ b/validation/longhorn/longhorn_test.go
@@ -41,11 +41,6 @@ type LonghornTestSuite struct {
 }
 
 func (l *LonghornTestSuite) TearDownSuite() {
-	l.T().Log("Uninstalling Longhorn chart")
-	err := charts.UninstallLonghornChart(l.client, charts.LonghornNamespace, l.cluster.ID, l.client.RancherConfig.Host)
-	if err != nil {
-		l.T().Logf("Failed to uninstall Longhorn chart: %v", err)
-	}
 	l.session.Cleanup()
 }
 
--- a/validation/longhorn/longhorn_test.go
+++ b/validation/longhorn/longhorn_test.go
@@ -136,6 +136,7 @@ func (l *LonghornTestSuite) TestScaleStatefulSetWithPVC() {
 
 	nodeList, err := steveClient.SteveType("node").List(nil)
 	require.NoError(l.T(), err)
+	require.NotEmpty(l.T(), nodeList.Data)
 
 	nodeCount := len(nodeList.Data)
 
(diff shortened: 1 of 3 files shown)


validation/longhorn/longhorn_test.go
```diff
@@ -199,14 +198,14 @@ func (l *LonghornTestSuite) TestScaleStatefulSetWithPVC() {
 	})
 	require.NoError(l.T(), err)
 
-	pods, err := steveClient.SteveType(shepherdPods.PodResourceSteveType).NamespacedSteveClient(namespace.Name).List(nil)
+	podList, err := steveClient.SteveType(shepherdPods.PodResourceSteveType).NamespacedSteveClient(namespace.Name).List(nil)
```
floatingman, line 201: Is the reason for these renamed variables to fix some kind of issue? Maybe with other variables with the same name?

validation/longhorn/longhorn_test.go
```diff
@@ -41,11 +41,6 @@ type LonghornTestSuite struct {
 }
 
 func (l *LonghornTestSuite) TearDownSuite() {
-	l.T().Log("Uninstalling Longhorn chart")
```
floatingman, line 44: Just curious why uninstalling Longhorn is failing?   I don't see this change as a problem, as we usually build new clusters with each test, but this probably will mess someone up who is running tests on a long-lived cluster.

floatingman (approved):

validation/longhorn/longhorn_test.go
lscalabrini01: The uninstall function keeps trying to uninstall Harvester indefinitely, waiting for a pod that never finishes. For this test, the cleanup function appears to be sufficient to uninstall Harvester

validation/longhorn/longhorn_test.go
lscalabrini01: No, it isn't. The renamed variable is just a minor refactor. I can revert it to the original variable name if you prefer

floatingman: It doesn't matter to me. I was just curious.

slickwarren (approved):

PR #725, opened 2026-06-16
Add tests for inherited namespace rules for downstream clusters

Priyashetty17:
Add tests for 2.15 RFE - Inherited namespace rules for downstream clusters
https://github.com/rancher/qa-tasks/issues/2352


joesims22 (approved):
lgtm

PR #728, opened 2026-06-17
Updating Appco fleet test

lscalabrini01:
This pull request updates the AppCo Istio validation flow and Jenkins individual E2E pipeline to support Application Collection (AppCo) authentication in tests, improve test logging, and adjust reporting/artifact handling.

Changes:

- Adds AppCo credential handling in the individual E2E Jenkins pipeline and writes them into the Docker env-file for test execution.
- Updates AppCo Istio tests to use logrus logging and simplifies/cleans up imports and assertions.
- Adds Fleet GitRepo helm repo URL regex configuration for AppCo OCI chart sources and improves AppCo flag help text.

### DIFF
--- a/validation/Jenkinsfile.individual.e2e
+++ b/validation/Jenkinsfile.individual.e2e
@@ -83,11 +83,6 @@ pipeline {
             defaultValue: '',
             description: 'Additional KEY=VALUE line to append to the .env file before running tests'
         )
-        booleanParam(
-            name: 'CAPTURE_IMAGES',
-            defaultValue: true,
-            description: 'When true, captures Docker images used during the test run'
-        )
         // ── Library parameter ────────────────────────────────────
         string(
             name: 'QA_JENKINS_LIBRARY_BRANCH',
--- a/validation/Jenkinsfile.individual.e2e
+++ b/validation/Jenkinsfile.individual.e2e
@@ -78,7 +78,16 @@ pipeline {
             defaultValue: 'build_qase_reporter_v2.sh',
             description: 'Qase reporter script name in validation/pipeline/scripts/'
         )
-
+        string(
+            name: 'ENV_VARIABLE',
+            defaultValue: '',
+            description: 'Additional KEY=VALUE line to append to the .env file before running tests'
+        )
+        booleanParam(
+            name: 'CAPTURE_IMAGES',
+            defaultValue: true,
+            description: 'When true, captures Docker images used during the test run'
+        )
         // ── Library parameter ────────────────────────────────────
         string(
             name: 'QA_JENKINS_LIBRARY_BRANCH',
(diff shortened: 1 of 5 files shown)


validation/Jenkinsfile.individual.e2e
```diff
@@ -78,7 +78,16 @@ pipeline {
...
+        string(
+            name: 'ENV_VARIABLE',
+            defaultValue: '',
+            description: 'Additional KEY=VALUE line to append to the .env file before running tests'
+        )
+        booleanParam(
+            name: 'CAPTURE_IMAGES',
+            defaultValue: true,
+            description: 'When true, captures Docker images used during the test run'
+        )
```
floatingman: It doesn't look like this is set up the same way as it is in validation/Jenkinsfile, so I don't think it's going to do anything.

validation/charts/appco/istio.go
```diff
@@ -21,6 +21,7 @@ import (
 const (
 	expectedDeployLog                                  = "deployed"
 	istioCanaryRevisionApp                             = "istiod-canary"
+	appCoRegistry                               string = `dp.apps.rancher.io`
```
floatingman, line 24: It doesn't look like this constant is used anywhere.

validation/charts/appco/istio_test.go
```diff
@@ -33,18 +32,18 @@ type IstioTestSuite struct {
 }
 
 func (i *IstioTestSuite) TearDownSuite() {
-	i.T().Log("Deleting Istio resources")
+	logrus.Info("Deleting Istio resources")
```
floatingman: Why did you change all of the logging? I think this might make it harder to debug because the log message won't be associated with the test case that caused it anymore.

floatingman (commented):
Just a few comments.

validation/charts/appco/istio.go
lscalabrini01: I forgot to replace the string, done!

validation/charts/appco/istio_test.go
lscalabrini01: We talked some time ago about using Logrus for logging. I'm fine with continuing to use the suite logger instead. Do you think we should keep using the suite logger rather than switching to Logrus?

validation/Jenkinsfile.individual.e2e
lscalabrini01: We should add the CAPTURE_IMAGE feature to this Jenkinsfile as well, but for now I'll remove this variable

validation/charts/appco/istio_test.go
floatingman: I don't recall why we wouldn't use the suite logger. When we don't, we lose details like this 
```
11:50:51      istio_test.go:195: Installing SideCar Istio AppCo
```
Where it will point to the line number in the test case.

lscalabrini01: done!

floatingman (approved):

PR #732, opened 2026-06-19
Update monitoring chart install to avoid waiting indefinitely on failure

lscalabrini01:
This PR updates the Rancher Monitoring chart install/upgrade wait logic to stop early when the chart transitions to a failed state, rather than continuing to wait until the watch times out.

Changes:

- Detect catalogv1.StatusFailed during Rancher Monitoring chart install and return an error with the chart’s reported failure reason.
- Detect catalogv1.StatusFailed during Rancher Monitoring chart upgrade (both “pending upgrade” and “fully deployed” waits) and return an error with the chart’s reported failure reason.


hamistao (approved):
Nice!

khushboo-rancher (approved):

PR #743, opened 2026-06-26
Fixing Appco fleet and neuvector failing tests

lscalabrini01:


hamistao (dismissed):

floatingman (dismissed):

khushboo-rancher (commented):
LGTM, qq to understand this - Why are we moving from token to admin-password?

floatingman (dismissed):
Looks good. Are you going to create a JJB PR as well?

floatingman:
> LGTM, qq to understand this - Why are we moving from token to admin-password?

@khushboo-rancher Token is also used; the bootstrap password is just used for the initial Rancher install, and the admin password is used if someone wants to log in and check the cluster manually. I think all of the tests still use a token to run.

floatingman (approved):

validation/Jenkinsfile.e2e
```diff
@@ -26,6 +26,8 @@
 def libraryBranch = env.QA_JENKINS_LIBRARY_BRANCH ?: 'main'
 library "qa-jenkins-library@${libraryBranch}"
 
+def DEFAULT_RANCHER_PASSWORD = 'XRi&Vfw8_Qr7*YVh1DE'
```
hamistao: what is this supposed to fix exactly?

and is it ok for us to include this in plain text on a public repo? (not sure but maybe it is)

floatingman: I agree. It's probably best to put this in the JJB repo.

lscalabrini01: The default parameter value in the declarative Jenkinsfile is overriding the default value configured in JJB

lscalabrini01: It is fixing the "Password must be at least 12 characters" rancher validation

hamistao: Oh so the problem here is that the default "rancherrocks" is invalid. It does have 12 characters, could it be that the problem is that it only contains small-case letters?

EIther way I agree with @floatingman and this should probably live over on JJB repo.

lscalabrini01: As I mentioned, the declarative Jenkins pipeline in rancher/tests overrides the parameters configured in the JJB repository

floatingman: @lscalabrini01 This password shouldn't be used in a public repository, because it is used elsewhere.

floatingman: Either change it to something that isn't used elsewhere or use another password.

floatingman (changes requested):

floatingman (approved):

khushboo-rancher (approved):

PR #744, opened 2026-06-30
bypass broken rancher2 provider signature via filesystem mirror

floatingman:
The rancher2 provider's release artifacts are signed by a key that does not match the signing key published in the OpenTofu registry (e.g. v13.1.4 is signed by 4CCE5F287729F2DFDDE1045F2EEB0F9AD44A135C but the registry advertises 6D5B085066648C32), so `tofu init` fails with "authentication signature from unknown issuer" before any apply/destroy can run. This is an upstream release defect in rancher/terraform-provider-rancher2 and affects every checked version (7.0.0, 13.1.4, 13.5.0). Note OPENTOFU_ENFORCE_GPG_VALIDATION=false does NOT bypass this: when the registry returns a key, OpenTofu always enforces (len(Keys) > 0 short-circuit in package_authentication.go).

Serve the provider from a filesystem mirror instead. The mirror path skips signature verification entirely (NewSignatureAuthentication is only called from registry_client.go), while the zip's SHA-256 is still checked against the registry's published shasum at image build time, and .terraform.lock.hcl still pins h1:/zh: hashes at runtime.

- Dockerfile.airgap-go-tests: add RANCHER2_PROVIDER_VERSION ARG; when set, fetch registry metadata, download the zip, verify its sha256, drop it into /usr/local/share/terraform/mirror, and write /root/.terraformrc so tofu serves rancher2 from the mirror while everything else stays on `direct`.
- Jenkinsfile.e2e: read the rancher2 version pinned in qa-infra-automation/tofu/rancher/cluster/.terraform.lock.hcl (checked out in the prior stage) and pass it as --build-arg so the mirror stays version-correct rather than hard-pinned.

Only rancher2 is mirrored; hashicorp/random and others retain full signature verification. Remove both blocks when upstream fixes provider signing.

### DIFF
--- a/validation/Jenkinsfile.e2e
+++ b/validation/Jenkinsfile.e2e
@@ -541,10 +541,31 @@ insecure = true'''
 
         stage('Build Docker Images') {
             steps {
-                // rancher-infra-tools:latest — used by make.groovy / infra.* for all make targets
-                sh "docker build --no-cache --platform linux/amd64 -t rancher-infra-tools:latest -f ${env.TESTS_DIR}/validation/pipeline/Dockerfile.infra ."
-                // rancher-go-test:latest — used for qase binary build and admin-token playbook
-                sh "docker build --no-cache --platform linux/amd64 -t rancher-go-test:latest -f ${env.TESTS_DIR}/validation/pipeline/Dockerfile.airgap-go-tests ."
+                script {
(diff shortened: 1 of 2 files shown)


floatingman:
Closing this issue as the issue is fixed upstream.
https://github.com/rancher/terraform-provider-rancher2/issues/2309

PR #748, opened 2026-07-02
Add GetGlobalSettingsNamesForCluster action for RA replacement test

joesims22:
The `TestRestrictedAdminReplacementListGlobalSettings` test was updated in https://github.com/rancher/tests/commit/71936e39782699f8bf1b8d08dac668e2c72e0ab4 to use Wrangler based shepherd extension to get the global settings. 

Previously the old helper listed settings through a cluster-scoped Wrangler context( the `/k8s/clusters/<id>` proxy), which the `restricted-admin-replacement` user can reach via it's inherited `cluster-owner` role.

The new Shepherd extension instead always uses the base Wrangler context (the Rancher management host `/apis/...)`, which the `restricted-admin-replacement` user isn't authorized against causing the test to fail with error: 
`failed to list settings:  (get settings.meta.k8s.io)`.

This PR adds an action that restores the cluster-scoped context and then delegates to the Shepherd extension.

### DIFF
--- a/actions/kubeapi/settings/settings.go
+++ b/actions/kubeapi/settings/settings.go
@@ -1,0 +1,27 @@
+package settings
+
+import (
+	"github.com/rancher/shepherd/clients/rancher"
+	
+	extclusterapi "github.com/rancher/shepherd/extensions/kubeapi/cluster"
+	extsettingsapi "github.com/rancher/shepherd/extensions/kubeapi/settings"
+)
+
+// GetGlobalSettingNamesForCluster lists global setting names using a cluster-scoped wrangler
+// context so non-admin users can read them via the downstream cluster proxy.
+func GetGlobalSettingNamesForCluster(client *rancher.Client, clusterID string) ([]string, error) {
+	settingsClient := client
+
+	if clusterID != extclusterapi.LocalCluster {
+		clusterContext, err := client.WranglerContext.DownStreamClusterWranglerContext(clusterID)
+		if err != nil {
+			return nil, err
+		}
+
+		scopedClient := *client
+		scopedClient.WranglerContext = clusterContext
+		settingsClient = &scopedClient
+	}
+
+	return extsettingsapi.GetGlobalSettingNames(settingsClient)
+}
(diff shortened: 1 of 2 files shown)


actions/kubeapi/settings/settings.go
```diff
@@ -0,0 +1,29 @@
...
+	
+	extclusterapi "github.com/rancher/shepherd/extensions/kubeapi/cluster"
+	extsettingsapi "github.com/rancher/shepherd/extensions/kubeapi/settings"
+)
+
+// GlobalSettingNamesForCluster returns the global setting names visible to the client. When the
+// target is a downstream cluster, the wrangler context is scoped to that cluster's proxy so that
+// non-admin users (e.g. restricted-admin replacements) can list settings via their inherited
+// cluster-owner permissions, then delegates to the shepherd settings extension.
+func GlobalSettingNamesForCluster(client *rancher.Client, clusterID string) ([]string, error) {
```
joesims22: The existing `GetGlobalSettingNames` Shepherd extension will not work with a non-admin user unless it's cluster-scoped.

Priyashetty17: nit, could we rename this to `GetGlobalSettingNamesForCluster`?

actions/kubeapi/settings/settings.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/rancher/shepherd/clients/rancher"
+	
+	extclusterapi "github.com/rancher/shepherd/extensions/kubeapi/cluster"
+	extsettingsapi "github.com/rancher/shepherd/extensions/kubeapi/settings"
+)
+
+// GlobalSettingNamesForCluster returns the global setting names visible to the client. When the
+// target is a downstream cluster, the wrangler context is scoped to that cluster's proxy so that
+// non-admin users (e.g. restricted-admin replacements) can list settings via their inherited
+// cluster-owner permissions, then delegates to the shepherd settings extension.
```
Priyashetty17: nit, The comment can be trimmed down, 
// GetGlobalSettingNamesForCluster lists global setting names using a cluster-scoped wrangler
// context so non-admin users can read them via the downstream cluster proxy.

Priyashetty17 (dismissed):
lgtm

Priyashetty17 (approved):

PR #751, opened 2026-07-07
Add `TestVolumeEncryption`

hamistao:
This PR introduces the test `TestVolumeEncryption` to our `LonghornTestSuite`. Among other small unrelated improvements to the codebase.

Said test required introducing new helper funcitions to handle longhorn resources such as voluemes, backups and backuptargets. These functions can also be moved to shepherd if we think it would make more sense for them to live there. Comments on the operations performed during the test are greatly appreciated as this is a very complex test and there are many possible approaches to satisfying the steps outlined in the pit schema for this test or possibly even improving the current schema.

So main topics that I am looking forward to getting feedback are:
1- Are the new helper functions fine here or should they be moved to shepherd?
2- Are the steps performed able to assure the features explored work as expected?
3- Since the test is quite big and only the second half of it requires interacting with S3, should we break this test into 2 or more tests?

This work performed as part of https://github.com/rancher/qa-tasks/issues/2403, the other test will be available for review in a separate PR.


actions/kubeapi/secrets/secrets.go
```diff
@@ -80,14 +80,14 @@ func CreateRegistrySecretDockerConfigJSON(registryconfig *Config) (string, error
 }
 
 // SecretCopyWithNewData is a helper to create a copy of an existing secret with new data.
-func SecretCopyWithNewData(secret *corev1.Secret, newData map[string][]byte) *corev1.Secret {
+func SecretCopyWithNewData(secret *corev1.Secret, newData map[string]string) *corev1.Secret {
 	updatedSecretObj := secret.DeepCopy()
 	if updatedSecretObj.Data == nil {
```
floatingman: Should this be changed to to ` if updatedSecretObj.StringData == nil`?

actions/kubeapi/longhorn/longhorn.go
```diff
@@ -0,0 +1,302 @@
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       longhornTypes.LonghornKindVolume,
+			APIVersion: longhorn.SchemeGroupVersion.Group + "/" + longhorn.SchemeGroupVersion.Version,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      namegenerator.AppendRandomString("restored-" + backup.Status.VolumeName),
+			Namespace: backup.Namespace,
+		},
+		Spec: longhorn.VolumeSpec{
+			Size:       backup.Spec.BackupBlockSize,
```
floatingman: Should this be the size stored in `backup.Status.VolumeSize` instead?  Seems like this will create a volume that is too small.

actions/kubeapi/longhorn/longhorn.go
```diff
@@ -0,0 +1,302 @@
...
+			Labels:       map[string]string{"type": "manual"},
+		},
+	}
+	err = backupController.Client().Create(context.Background(), namespace, backup, backup, metav1.CreateOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("Error creating backup from snapshot %s on namespace %s: %w", snapshot.Name, namespace, err)
+	}
+
+	client.Session.RegisterCleanupFunc(func() error {
+		return snapshotController.Client().Delete(context.Background(), charts.LonghornNamespace, snapshot.Name, metav1.DeleteOptions{})
```
floatingman: Should this be `backupController` instead of `snapshotController`?

actions/kubeapi/longhorn/longhorn.go
```diff
@@ -0,0 +1,302 @@
...
+		Group:   longhorn.SchemeGroupVersion.Group,
+		Version: longhorn.SchemeGroupVersion.Version,
+		Kind:    longhornTypes.LonghornKindVolume,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var volume longhorn.Volume
+	err = volumeController.Client().Get(context.Background(), namespace, volumeName, &volume, metav1.GetOptions{})
```
floatingman, line 132: I noticed that some of the methods use `namespace` and then some use `charts.LonghornNamespace` . Which is the correct one to use?

actions/storage/s3/s3.go
```diff
@@ -51,6 +56,62 @@ func CreateS3Bucket(bucketName, region, accessKey, secretKey string) error {
...
+	cfg, err := awsconfig.LoadDefaultConfig(context.Background(), awsconfig.WithRegion(region), awsconfig.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(awsCreds.AccessKey, awsCreds.SecretKey, "")))
+	if err != nil {
+		return nil, err
+	}
+
+	s3Client := s3.NewFromConfig(cfg)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+	thing, err := manager.New(s3Client).GetObject(ctx, &transfermanager.GetObjectInput{
```
floatingman: Two things.  
1. It's a good idea to add a `defer thing.Body.Close() after this call to prevent resource leaks.
2. The error is never checked/handled for this call.

actions/storage/s3/s3.go
```diff
@@ -1,15 +1,20 @@
...
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awsconfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager"
+	manager "github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager"
```
floatingman: These 2 imports are the same thing. You should just keep one.

floatingman (commented):
Just a few comments, but I think the one in secrets.go might cause a panic if not handled.

actions/kubeapi/longhorn/longhorn.go
hamistao: Good catch! fixed

actions/kubeapi/longhorn/longhorn.go
hamistao: The methods that take in a `namespace` parameter receive `charts.LonghornNamespace` so I hadn't noticed the inconsistency. This is now fixed and now `CreateS3LonghornBackupTarget` always use `charts.LonghornNamespace` since the target namespace matters very little, `CreateLonghornVolumeBackup` takes in the volume's namespace and creates the backup in the same namespace and `RestoreLonghornVolumeFromBackup` creates the restored volume in the same namespace as the backup for simplicity. In the end we always expect the value to be `charts.LonghornNamespace` since it is the approriate place for longhorn resources to live.

actions/storage/s3/s3.go
hamistao: Added the error checking. `thing.Body` is an `io.Reader`, not an `io.ReadCloser`, so it doesn't have a `Close` method.

actions/storage/s3/s3.go
hamistao: Done

actions/kubeapi/secrets/secrets.go
hamistao: Yep

actions/kubeapi/longhorn/longhorn.go
hamistao: You are probably correct, I will chase the best way to handle this.

actions/charts/ranchermonitoring.go
```diff
@@ -139,7 +141,7 @@ func InstallRancherMonitoringChart(client *rancher.Client, installOptions *Insta
 			return true, nil
 		}
 		if state == string(catalogv1.StatusFailed) {
-			return false, fmt.Errorf("monitoring chart installation failed: %s", app.Status.Summary.Error)
+			return false, fmt.Errorf("monitoring chart installation failed: %t", app.Status.Summary.Error)
```
lscalabrini01: Is the `app.Status.Summary.Error` a boolean? if not I would recommend use `%s`

validation/longhorn/longhorn_test.go
```diff
@@ -232,6 +245,198 @@ func (l *LonghornTestSuite) TestScaleStatefulSetWithPVC() {
...
+	nginxSpec := &appv1.DeploymentSpec{}
+	err = steveV1.ConvertToK8sType(nginxResponse.Spec, nginxSpec)
+	require.NoError(l.T(), err)
+	require.Equal(l.T(), len(nginxSpec.Template.Spec.Volumes), 1)
+
+	volumeName := nginxSpec.Template.Spec.Volumes[0].Name
+
+	l.client.Session.RegisterCleanupFunc(func() error {
+		return longhornActions.DeleteLonghornVolume(l.client, l.cluster.ID, charts.LonghornNamespace, volumeName)
+	})
```
lscalabrini01: I would recommend moving the cleanup function to the actions package instead of keeping it in the test. That way, it can be reused by any other test that needs to create a volume

actions/kubeapi/longhorn/longhorn.go
```diff
@@ -0,0 +1,302 @@
...
+
+	client.Session.RegisterCleanupFunc(func() error {
+		return backupTargetController.Client().Delete(context.Background(), charts.LonghornNamespace, targetName, metav1.DeleteOptions{})
+	})
+
+	return backupTarget, nil
+}
+
+// CreateLonghornVolumeBackup creates a snapshot for a Longhorn Volume and then creates a backup from this snapshot.
+// This waits until the backuṕ is succesfully created and returns the URL for that backup in S3.
```
lscalabrini01: Suggestion: from `backuṕ` to `backup`

actions/storage/pvc.go
```diff
@@ -21,16 +23,48 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
lscalabrini01: Duplicate import

lscalabrini01 (changes requested):

actions/charts/ranchermonitoring.go
hamistao: It is a boolean, but I failed to realize at the time that there is no use logging this value like this, I changed it up to:

`return false, fmt.Errorf("monitoring chart installation failed with %s status", app.Status.Summary.State)`

Not very imformative but definetely an improvement.

actions/kubeapi/longhorn/longhorn.go
hamistao: Ok so I just had to do some conversions on `backup.Status.VolumeSize` since it is a string but otherwise this works just fine.

validation/longhorn/longhorn_test.go
hamistao: OK so I had to check if the volume is a longhorn on creation since this cleanup step is only needed for longhorn volumes. That ended up working great and I liked the result.

lscalabrini01 (dismissed):

floatingman (approved):

lscalabrini01 (approved):

slickwarren (approved):

PR #758, opened 2026-07-13
 Modified observedGenererationLocal/Remote value for CRTB Status field test

joesims22:
Checks on the observedGenerationLocal/Remotehave value been modified to handle at least one reconciliation occurring to the CRTB status. This should eliminate flaky tests that were failing due to reconciliation timing.


Priyashetty17 (approved):
lgtm

PR #759, opened 2026-07-13
add security group cleanup to custodian

slickwarren:


scripts/custodian/aws.yaml
```diff
@@ -497,3 +497,36 @@ policies:
...
+  filters:
+    - 'tag:DoNotDelete': absent
+    - 'tag:doNotDelete': absent
+    - type: unused
+    - GroupName: {op: ne, value: default}
+    - type: marked-for-op
+      tag: ec2_known_user
+      op: delete
+  actions:
+    - delete
```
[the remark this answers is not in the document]
slickwarren: the docs clearly say this is allowed and expected for this resource: https://cloudcustodian.io/docs/aws/resources/security-group.html#aws-security-group-actions-delete

floatingman (approved):

scripts/custodian/aws.yaml
```diff
@@ -497,3 +497,36 @@ policies:
...
     - type: delete
+
+# Security Groups
+- name: schedule-delete-securitygroup
+  resource: security-group
+  description: |
+    Mark unused, non-default security groups for deletion.
+  filters:
+    - 'tag:DoNotDelete': absent
+    - 'tag:doNotDelete': absent
```
hamistao, line 508: Same filter twice?

scripts/custodian/aws.yaml
```diff
@@ -497,3 +497,36 @@ policies:
...
+    - 'tag:ec2_known_user': absent
```
hamistao, line 509: can you please elaborate on what this tag does?

scripts/custodian/aws.yaml
```diff
@@ -497,3 +497,36 @@ policies:
       op: eq
   actions:
     - type: delete
+
+# Security Groups
+- name: schedule-delete-securitygroup
+  resource: security-group
+  description: |
+    Mark unused, non-default security groups for deletion.
+  filters:
```
hamistao, line 506: Not that it matters much but would it be possible to filter for resources that are already marked?

khushboo-rancher (commented):
Thanks!
Overall looks good. Just couple of questions:
1. I'm not sure about nested operators support like `{op: ne, value: default}`. Do we want to make it consistent with file and use a standard `type: value`?
```
- type: value
      key: GroupName
      op: ne
      value: default
```
2. nit - Do we want to make `action: delete` consistent with the file as below?
```
actions:
    - type: delete
```

scripts/custodian/aws.yaml
```diff
@@ -497,3 +497,36 @@ policies:
...
+    - 'tag:DoNotDelete': absent
+    - 'tag:doNotDelete': absent
+    - 'tag:ec2_known_user': absent
+    - type: unused
+    - GroupName: {op: ne, value: default}
+  actions:
+    - type: mark-for-op
+      tag: ec2_known_user
+      op: delete
+      days: 3
```
lscalabrini01, line 516: The ec2 instances use to be deleted in 2 days, could we change this value for 2?

lscalabrini01 (approved):

PR #762, opened 2026-07-14
Fix compare rancher tag action to have latest-tag-v215 input

markusewalker:
This PR updates the compare rancher tag action to use the latest-tag-v215 input. This ensures that the action references the intended tag version for comparison.


markusewalker:
As mentioned offline, this PR is addressing a release automation issue. Check Rancher Tag will run soon, so will merge once the CI passes.

PR #775, opened 2026-07-22
Align tests and schemas

hamistao:
As part of a personal Goal, I experimented with using AI to identify discrepancies between our automated tests and their schemas and this PR is a result of trying to fix the identified discrepancies.

The more convoluted fixes here would be:
- The `helm install` commands from `TestAmbientInstallation` and `TestGatewayStandaloneInstallation` schemas were updated to match the one used on the tests.
- `TestLoadBalancer` and `TestLoadBalanceScaleAndUpgrade` were updated to test the connectivity to the actual Load Balancers instead of the Node port

### DIFF
--- a/actions/networking/verify.go
+++ b/actions/networking/verify.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
-	"strconv"
 	"strings"
 	"testing"
 
@@ -88,6 +87,66 @@ func VerifyNetworkPolicy(client *rancher.Client, clusterID string, namespaceName
 	return nil
 }
 
+// verifyConnectivityFromWorkerNodes verifies if any worker node in the cluster is able to access the provided ip:port
+// and retrieve the expected content from name.html.
+func verifyConnectivityFromWorkerNodes(client *rancher.Client, clusterID string, ip string, port int, workloadName string) error {
+	query, err := url.ParseQuery(clusters.LabelWorker)
+	if err != nil {
+		return err
+	}
+
+	steveClient, err := client.Steve.ProxyDownstream(clusterID)
+	if err != nil {
+		return err
+	}
+
+	nodeList, err := steveClient.SteveType(stevetypes.Node).List(query)
+	if err != nil {
+		return err
+	}
+
+	if len(nodeList.Data) == 0 {
+		return errors.New("no worker nodes found")
+	}
+
+	for _, machine := range nodeList.Data {
+		sshNode, err := sshkeys.GetSSHNodeFromMachine(client, &machine)
+		if err != nil {
+			logrus.Debugf("Could not SSH into worker node %s: %s", machine.Name, err.Error())
+			continue
+		}
+
+		logrus.Debugf("Curling '%s:%d/name.html' from node %s", ip, port, machine.Name)
+		log, err := sshNode.ExecuteCommand(fmt.Sprintf("curl -s %s:%d/name.html", ip, port))
+		if err != nil && !errors.Is(err, &ssh.ExitMissingError{}) {
+			logrus.Debugf("Curl failed on node %s: %v", machine.Name, err)
+			continue
+		}
+
+		if strings.Contains(log, workloadName) { // This should be one of the pod's names.
+			return nil
+		} else {
+			logrus.Debugf("Curl result %s doesn't contain expected content '%s'", log, workloadName)
+		}
+	}
+
+	return fmt.Errorf("Unable to connect to %s:%d/name.html from any worker node", ip, port)
+}
+
+func verifyConnectivityFromPod(client *rancher.Client, clusterID string, ip string, port int, workloadName string) error {
+	execCmd := []string{"curl", "-s", fmt.Sprintf("%s:%d/name.html", ip, port)}
+	log, err := kubectl.Command(client, nil, clusterID, execCmd, "")
+	if err != nil {
+		return err
+	}
+
+	if !strings.Contains(log, workloadName) { // This should be one of the pod's names.
+		return fmt.Errorf("Curl result %s doesn't include the workload name %s", log, workloadName)
+	}
+
+	return nil
+}
+
 // VerifyNodePortConnectivity verifies that the node port is accessible by curling the worker node external IP
 func VerifyNodePortConnectivity(client *rancher.Client, clusterID string, nodePort int, workloadName string) error {
 	steveClient, err := client.Steve.ProxyDownstream(clusterID)
@@ -122,21 +181,14 @@ func VerifyNodePortConnectivity(client *rancher.Client, clusterID string, nodePo
 		}
 
 		logrus.Debugf("Curling node port %d on node %s (%s)", nodePort, machine.Name, nodeIP)
-		execCmd := []string{"curl", fmt.Sprintf("%s:%s/name.html", nodeIP, strconv.Itoa(nodePort))}
-		log, err := kubectl.Command(client, nil, clusterID, execCmd, "")
-		if err != nil {
-			return fmt.Errorf("curl command failed on node %s: %w", machine.Name, err)
-		}
-
-		if strings.Contains(log, workloadName) {
-			return nil
-		}
+		verifyConnectivityFromPod(client, clusterID, nodeIP, nodePort, workloadName)
 	}
 
 	return fmt.Errorf("unable to access node port %d for workload %s", nodePort, workloadName)
 }
 
 // VerifyLoadBalancerConnectivity verifies that the Load Balancer service is accessible by curling its IP:port.
+// This includes
 func VerifyLoadBalancerConnectivity(t *testing.T, client *rancher.Client, clusterID string, serviceID string, workloadName string) {
 	steveClient, err := client.Steve.ProxyDownstream(clusterID)
 	require.NoError(t, err)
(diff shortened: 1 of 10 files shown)


actions/networking/verify.go
```diff
@@ -6,15 +6,18 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"testing"
 
 	"github.com/rancher/shepherd/clients/rancher"
+	steveV1 "github.com/rancher/shepherd/clients/rancher/v1"
 	v1 "github.com/rancher/shepherd/clients/rancher/v1"
```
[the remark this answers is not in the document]
lscalabrini01: It's a good pointing

actions/networking/verify.go
```diff
@@ -134,6 +137,32 @@ func VerifyNodePortConnectivity(client *rancher.Client, clusterID string, nodePo
...
+func VerifyLoadBalancerConnectivity(t *testing.T, client *rancher.Client, clusterID string, serviceID string, workloadName string) {
+	steveClient, err := client.Steve.ProxyDownstream(clusterID)
+	require.NoError(t, err)
+
+	service, err := steveClient.SteveType(stevetypes.Service).ByID(serviceID)
+	require.NoError(t, err)
+
+	k8sService := &corev1.Service{}
+	err = steveV1.ConvertToK8sType(service, k8sService)
+	require.NoError(t, err)
```
lscalabrini01: We usually just return the error from the action and verify it in the test. As an example: https://github.com/hamistao/tests/blob/a1642af8087366191e6d61b758e7cf7ea600028e/actions/networking/verify.go#L49

lscalabrini01 (changes requested):

actions/networking/verify.go
hamistao: Are you suggesting that I remove `t *testing.T` from the parameters? I don't think it makes sense to take in the parameter and not use it here. The example you gave doesn't include such a parameter thus it makes sense for it to return the error.

I decided to include this parameter here so we could have the IP and port in the logging as well, which I think is nice.

hamistao: I can use logrus here instead if you prefer

lscalabrini01: I haven't seen the `t *testing.T`, yes you can remove it and use the logrus

actions/networking/verify.go
```diff
@@ -86,56 +85,68 @@ func VerifyNetworkPolicy(client *rancher.Client, clusterID string, namespaceName
...
+
+	if !strings.Contains(log, workloadName) { // This should be one of the pod's names.
+		return fmt.Errorf("Curl result %s doesn't include the workload name %s", log, workloadName)
+	}
+
+	return nil
+}
+
+// VerifyNodePortConnectivity verifies that the node port is accessible by curling the worker node external IP
+func VerifyNodePortConnectivity(client *rancher.Client, clusterID string, nodePort int, workloadName string) error {
```
slickwarren: VerifyNodePortConnectivity can never succeed
The return value of verifyConnectivityFromPod is ignored:

verifyConnectivityFromPod(client, clusterID, nodeIP, nodePort, workloadName)
Because the result is discarded, the loop always falls through to the final return fmt.Errorf("unable to access node port ..."), so this function can never return success. This also leaves an unchecked error (lint/vet issue). It should be:

if err := verifyConnectivityFromPod(client, clusterID, nodeIP, nodePort, workloadName); err == nil {
    return nil
}

actions/networking/verify.go
```diff
@@ -156,78 +167,80 @@ func VerifyHostPortConnectivity(client *rancher.Client, clusterID string, hostPo
...
-	serviceResp, err := steveClient.SteveType(stevetypes.Service).ByID(serviceID)
+	service, err := steveClient.SteveType(stevetypes.Service).ByID(serviceID)
 	if err != nil {
 		return err
 	}
 
-	newService := &corev1.Service{}
-	err = v1.ConvertToK8sType(serviceResp.JSONResp, newService)
+	k8sService := &corev1.Service{}
+	err = v1.ConvertToK8sType(service, k8sService)
```
slickwarren: Wrong argument passed to ConvertToK8sType in VerifyLoadBalancerConnectivity — actions/networking/verify.go

service, err := steveClient.SteveType(stevetypes.Service).ByID(serviceID)
...
err = v1.ConvertToK8sType(service, k8sService)
This passes the raw Steve object instead of service.JSONResp. Every other call site (including VerifyClusterConnectivity in this same file) passes .JSONResp, and conversion will likely fail at runtime. Should be:

err = v1.ConvertToK8sType(service.JSONResp, k8sService)

validation/networking/connectivity/port_test.go
```diff
@@ -215,68 +214,58 @@ func (p *PortTestSuite) TestClusterIP() {
...
+	require.NoError(p.T(), err)
 
-			logrus.Infof("Verifying load balancer connectivity for daemonset %s on node port %d", testDaemonset.Name, nodePort)
-			err = networking.VerifyNodePortConnectivity(p.client, p.cluster.ID, nodePort, testDaemonset.Name)
-			require.NoError(p.T(), err)
-		})
-	}
+	err = networking.VerifyLoadBalancerConnectivity(p.client, p.cluster.ID, serviceResp.ID, testDaemonset.Name)
+	require.NoError(p.T(), err)
+	require.NoError(p.T(), err)
```
slickwarren: duplicate assertion here

slickwarren (changes requested):
I think we still need changes here; can you send a link to a test run using this? 

Lastly can you confirm all callers of the changed signatures are updated:

VerifyClusterConnectivity changed its parameter from path string to port int. Please confirm there are no other callers of VerifyClusterConnectivity or VerifyNodePortConnectivity elsewhere in the repo that still use the old signature.

actions/networking/verify.go
hamistao: I believe `ConvertToK8sType` is flexible enough so that it works with both objects (it takes an `any` parameter). But I changed this as you suggested either way to keep this more standardized.

actions/networking/verify.go
hamistao: My bad, I failed to push just these changes before reaching out to you guys. Fixed.

lscalabrini01 (dismissed):

actions/networking/verify.go
```diff
@@ -86,98 +85,115 @@ func VerifyNetworkPolicy(client *rancher.Client, clusterID string, namespaceName
...
+			newNode := &corev1.Node{}
+			err = v1.ConvertToK8sType(machine.JSONResp, newNode)
+			if err != nil {
+				return fmt.Errorf("failed to convert node %s: %w", machine.Name, err)
+			}
+
+			ip = kubeapinodes.GetNodeIP(newNode, corev1.NodeExternalIP)
+			if ip == "" {
+				ip = kubeapinodes.GetNodeIP(newNode, corev1.NodeInternalIP)
+			}
```
floatingman: One thing I noticed: when something calls this function and passes ip = "", only the first node's IP that can be SSHed into is used, and then that IP is used for all subsequent nodes. So tests in TestNodePort and TestNodePortScaleAndUpgrade could pass even if node 2 is broken.
It's a better idea to use a local variable instead of mutating the ip variable. Something like 
```
nodeIP := ip
if nodeIP == "" {
    newNode := &corev1.Node{}
    ...
    nodeIP = kubeapinodes.GetNodeIP(newNode, corev1.NodeExternalIP)
    if nodeIP == "" {
        nodeIP = kubeapinodes.GetNodeIP(newNode, corev1.NodeInternalIP)
    }
}
```

actions/networking/verify.go
```diff
@@ -86,98 +85,115 @@ func VerifyNetworkPolicy(client *rancher.Client, clusterID string, namespaceName
...
-		if strings.Contains(output, workloadName) {
-			return nil
-		}
+	if len(k8sService.Status.LoadBalancer.Ingress) == 0 {
+		return fmt.Errorf("No ingress (Status.LoadBalancer.Ingress) specified in service %s", k8sService.Name)
 	}
 
-	return fmt.Errorf("unable to access host port %d for workload %s on any worker node", hostPort, workloadName)
+	port := k8sService.Spec.Ports[0].Port
+	ip := k8sService.Status.LoadBalancer.Ingress[0].IP
```
floatingman: It's possible this will be empty for cloud providers that expose the load balancer by DNS name rather than IP.
You can add a check for Hostname to get around this issue.
```
ingress := k8sService.Status.LoadBalancer.Ingress[0]
ip := ingress.IP
if ip == "" {
    ip = ingress.Hostname
}
```

actions/networking/verify.go
```diff
@@ -86,98 +85,115 @@ func VerifyNetworkPolicy(client *rancher.Client, clusterID string, namespaceName
...
+	if !strings.Contains(log, workloadName) { // This should be one of the pod's names.
+		return fmt.Errorf("Curl result %s doesn't include the workload name %s", log, workloadName)
 	}
 
-	nodeList, err := steveClient.SteveType(stevetypes.Node).List(query)
+	return nil
+}
+
+// VerifyLoadBalancerConnectivity verifies that the Load Balancer service is accessible by curling its IP:port.
+// This includes
```
floatingman: I think this comment got cut off.

slickwarren (approved):

floatingman (approved):

lscalabrini01 (approved):

PR #794, opened 2026-08-03
Add principal search coverage to OpenLDAP and Active Directory suites

dasarinaidu:
## Problem

Both auth provider suites hand-construct principal IDs via `GetGroupPrincipalID` / `GetUserPrincipalID` and never call `POST /v3/principals?action=search`. A pure findability regression in the search endpoint — the one the UI uses to populate every principal picker — leaves the whole suite green.

## What this adds

Eight tests per provider (OpenLDAP and Active Directory) that assert the principal ID returned by search matches the constructed one, compared case-insensitively since DNs are not case-sensitive.

| Test | Catches |
|---|---|
| `…PrincipalSearchFindsGroups` | Groups unfindable by exact name, across nesting levels |
| `…PrincipalSearchFindsUsers` | Users unfindable by exact name, across nesting levels |
| `…PrincipalSearchFindsLocalUserByDisplayName` | A local user no longer matching on display name |
| `…PrincipalSearchProvisionedUserIsNotLocal` | An externally provisioned user surfacing as a local principal |
| `…PrincipalSearchByPartialName` | Prefix/substring matching regressions — the UI typeahead path |
| `…PrincipalSearchByPrincipalType` | A broken `principalType` filter returning off-type results |
| `…PrincipalByIDResolvesGroupAndUser` | `/v3/principals/<id>` failing to resolve provider, type or display name |
| `…PrincipalSearchAfterProviderDisabled` | Principals still served from a disabled provider |

## Context on the `IsNotLocal` tests

Related to rancher/rancher#56362. The reporter's stated root cause there does not hold — rancher/rancher#56081 changed `isLocalUser`, which filters results rather than changing what matches, and the display-name index still matches. #56081 (deduplicating external users) and #56362 want opposite behaviour, so only one can hold.

`…PrincipalSearchProvisionedUserIsNotLocal` exercises the actual discriminator: a provisioned user has an empty `username` and more than one principal ID. If `isLocalUser` is later relaxed to satisfy #56362, these tests go red across AD, OpenLDAP and the existing SCIM suite at the same time, which makes the trade-off visible rather than silent.

## Helper placement

Four new helpers in `actions/auth/verify.go`, all Action-tier per the [Actions vs Extensions](https://confluence.suse.com/spaces/RANQA/pages/1505559041/Actions+vs+Extensions) rubric — they are `Verify`-prefixed test-flow assertions with no CRUD, and they sit alongside the existing `VerifyPrincipalIsLocal` / `VerifyPrincipalNotLocal`:

- `VerifyPrincipalSearchReturnsID`
- `VerifyPrincipalSearchByTypeReturnsOnly`
- `VerifyPrincipalSearchExcludesProvider`
- `VerifyPrincipalByID`

The shared `searchPrincipals` helper takes a `principalType` (empty for unrestricted). `NewPrincipalID`, `GetGroupPrincipalID` and `GetUserPrincipalID` now use new `PrincipalTypeUser` / `PrincipalTypeGroup` constants instead of string literals.

No shepherd change is required.

## Notes for reviewers

- `…PrincipalSearchAfterProviderDisabled` mutates cluster state mid-suite. It uses the local admin client rather than the LDAP-authenticated one for the post-disable search, because disabling the provider invalidates the LDAP-issued token, and it re-enables the provider at the end.
- `VerifyPrincipalByID` path-escapes the ID: norman's `DoByID` concatenates it into the URL with no escaping, and a principal ID embeds both a scheme separator and a DN. There was no prior use of `Principal.ByID` in this repo or shepherd; it has been confirmed working against a live setup.

## Testing

All sixteen tests pass against a live Rancher with both providers configured. `go vet -tags validation ./validation/auth/provider/...` and `go build ./...` in the `actions` module are clean.

### DIFF
--- a/actions/auth/authprovider.go
+++ b/actions/auth/authprovider.go
@@ -29,6 +29,10 @@ const (
 	ActiveDirectory                      = "activedirectory"
 	OpenLdapPasswordSecretID             = "openldapconfig-serviceaccountpassword"
 	ActiveDirectoryPasswordSecretID      = "activedirectoryconfig-serviceaccountpassword"
+	PrincipalTypeUser                    = "user"
+	PrincipalTypeGroup                   = "group"
+	AccessModeMissingRequiredError       = "accessMode=MissingRequired"
+	PermissionDeniedError                = "PermissionDenied"
 )
 
 type User struct {
@@ -107,7 +111,7 @@ func LoginAsAuthUser(client *rancher.Client, user *v3.User, providerName string)
 func NewPrincipalID(authConfigID, principalType, name, userSearchBase, groupSearchBase string) string {
 	baseDN := userSearchBase
 
-	if principalType == "group" {
+	if principalType == PrincipalTypeGroup {
 		baseDN = groupSearchBase
 	}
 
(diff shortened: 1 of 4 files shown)


joesims22 (approved):
lgtm

PR #795, opened 2026-08-04
Add `TestMonitoringIntegration`

hamistao:
Implementing this test clearly required a schema update since:
1- The old schema relied heavily on the use of UI;
2- The set up of Longhorn metrics on Grafana was innacurate;

So this is my attempt at improving it, running Prometheus queries with the Grafana API while using Rancher as a proxy to check the values shown in Grafana without relying on the UI. 

Part of https://github.com/rancher/qa-tasks/issues/2403

### DIFF
--- a/actions/monitoring/grafana.go
+++ b/actions/monitoring/grafana.go
@@ -18,7 +18,6 @@ import (
 
 const (
 	GrafanaProxyURLTemplate = "https://%s/k8s/clusters/%s/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy"
-	LoginURL                = "/login"
 	DsQueryURL              = "/api/ds/query"
 )
 
@@ -35,11 +34,18 @@ type QueryResponse struct {
 }
 
 func doRequest(client http.Client, method string, url string, payload any) ([]byte, []*http.Cookie, error) {
-	req, _ := http.NewRequest(method, url, nil)
+	req, err := http.NewRequest(method, url, nil)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	if payload != nil {
 		req.Header.Set("Content-Type", "application/json")
-		jsonPayload, _ := json.Marshal(payload)
+		jsonPayload, err := json.Marshal(payload)
+		if err != nil {
+			return nil, nil, fmt.Errorf("payload is not proper JSON: %w", err)
+		}
+
 		req.Body = io.NopCloser(bytes.NewBuffer(jsonPayload))
 	}
 
@@ -48,17 +54,21 @@ func doRequest(client http.Client, method string, url string, payload any) ([]by
 		return nil, nil, err
 	}
 	defer resp.Body.Close()
-	body, _ := io.ReadAll(resp.Body)
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusCreated {
 		return body, resp.Cookies(), nil
 	}
 
-	return nil, nil, fmt.Errorf("Request to %s failed with %d: %s\n", url, resp.StatusCode, string(body))
+	return nil, nil, fmt.Errorf("request to %s failed with %d: %s", url, resp.StatusCode, string(body))
 }
 
-// LoginGrafana logs in to a Rancher-proxied Grafana and returns a client embedded with a cookie containing a Ranher user token.
-func LoginGrafana(rancherHost string, rancherToken string, clusterID string, skipTLS bool) (*http.Client, error) {
+// NewGrafanaProxyClient creates Rancher-proxied client to Grafana by embedding it with a cookie containing a Rancher user token and returns it.
+func NewGrafanaProxyClient(rancherHost string, rancherToken string, skipTLS bool) (*http.Client, error) {
 	jar, _ := cookiejar.New(nil)
 	httpClient := &http.Client{
 		Transport: &http.Transport{
@@ -104,7 +114,7 @@ func PrometheusQueryInGrafana(client *http.Client, rancherHost string, clusterID
 		}
 
 		if err := json.Unmarshal(respBytes, &resp); err != nil {
-			return false, err
+			return false, nil
 		}
 
 		if len(resp.Results.A.Frames) == 0 {
--- a/actions/monitoring/grafana.go
+++ b/actions/monitoring/grafana.go
@@ -1,0 +1,125 @@
+package monitoring
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/cookiejar"
+	"net/url"
+	"time"
+
+	"github.com/rancher/shepherd/extensions/defaults"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	GrafanaProxyURLTemplate = "https://%s/k8s/clusters/%s/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy"
+	LoginURL                = "/login"
+	DsQueryURL              = "/api/ds/query"
+)
+
+type QueryResponse struct {
+	Results struct {
+		A struct {
+			Frames []struct {
+				Data struct {
+					Values [][]int `json:"values"`
+				} `json:"data"`
+			} `json:"frames"`
+		} `json:"A"`
+	} `json:"results"`
+}
+
+func doRequest(client http.Client, method string, url string, payload any) ([]byte, []*http.Cookie, error) {
+	req, _ := http.NewRequest(method, url, nil)
+
+	if payload != nil {
+		req.Header.Set("Content-Type", "application/json")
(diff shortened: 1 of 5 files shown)


validation/longhorn/longhorn_test.go
```diff
@@ -10,13 +10,16 @@ import (
...
 	"testing"
 
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/rancher/norman/types"
 	"github.com/rancher/shepherd/clients/rancher"
 	"github.com/rancher/shepherd/clients/rancher/catalog"
 	management "github.com/rancher/shepherd/clients/rancher/generated/management/v3"
 	steveV1 "github.com/rancher/shepherd/clients/rancher/v1"
+	extencharts "github.com/rancher/shepherd/extensions/charts"
 	shepherdCharts "github.com/rancher/shepherd/extensions/charts"
```
floatingman: This is imported twice.

actions/monitoring/grafana.go
```diff
@@ -0,0 +1,125 @@
...
+				Data struct {
+					Values [][]int `json:"values"`
+				} `json:"data"`
+			} `json:"frames"`
+		} `json:"A"`
+	} `json:"results"`
+}
+
+func doRequest(client http.Client, method string, url string, payload any) ([]byte, []*http.Cookie, error) {
+	req, _ := http.NewRequest(method, url, nil)
```
floatingman: You should handle an error if NewRequest fails. This could cause a panic later on.

actions/monitoring/grafana.go
```diff
@@ -0,0 +1,125 @@
...
+		jsonPayload, _ := json.Marshal(payload)
+		req.Body = io.NopCloser(bytes.NewBuffer(jsonPayload))
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer resp.Body.Close()
+	body, _ := io.ReadAll(resp.Body)
```
floatingman: You should handle the ReadAll error. It can cause issues when the body variable is used later.

actions/monitoring/grafana.go
```diff
@@ -0,0 +1,125 @@
...
+		} `json:"A"`
+	} `json:"results"`
+}
+
+func doRequest(client http.Client, method string, url string, payload any) ([]byte, []*http.Cookie, error) {
+	req, _ := http.NewRequest(method, url, nil)
+
+	if payload != nil {
+		req.Header.Set("Content-Type", "application/json")
+		jsonPayload, _ := json.Marshal(payload)
```
floatingman: You should handle the errors from Marshal. This can make it hard to debug.

actions/monitoring/grafana.go
```diff
@@ -0,0 +1,125 @@
...
+			return false, err
+		}
+
+		if err := json.Unmarshal(respBytes, &resp); err != nil {
+			return false, err
+		}
+
+		if len(resp.Results.A.Frames) == 0 {
+			return false, fmt.Errorf("No frames returned")
+		}
```
floatingman, line 122: Returning non-nil errors will cause PollUntilContextTimeout to abort immediately and won't retry for 30 seconds. If you want to retry on these errors, then `return false, nil` for each case that you want to retry and only return errors for truly unrecoverable errors.

actions/monitoring/grafana.go
```diff
@@ -0,0 +1,125 @@
...
+	parsedBaseURL, err := url.Parse("https://" + rancherHost)
+	if err != nil {
+		return nil, err
+	}
+
+	cookies := []*http.Cookie{{Name: "R_SESS", Value: rancherToken}}
+	httpClient.Jar.SetCookies(parsedBaseURL, cookies)
+
+	return httpClient, nil
+}
```
floatingman: There's a typo in the comment for this function. Also, clusterID is not used. This function doesn't really log in; it creates a client. It really should be named something like NewGrafanaProxyClient.

actions/monitoring/grafana.go
```diff
@@ -0,0 +1,125 @@
...
+	"net/url"
+	"time"
+
+	"github.com/rancher/shepherd/extensions/defaults"
+	kwait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	GrafanaProxyURLTemplate = "https://%s/k8s/clusters/%s/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy"
+	LoginURL                = "/login"
```
floatingman: It doesn't seem like this constant is used anywhere.

actions/monitoring/grafana.go
```diff
@@ -0,0 +1,125 @@
...
+			Frames []struct {
+				Data struct {
+					Values [][]int `json:"values"`
+				} `json:"data"`
+			} `json:"frames"`
+		} `json:"A"`
+	} `json:"results"`
+}
+
+func doRequest(client http.Client, method string, url string, payload any) ([]byte, []*http.Cookie, error) {
```
floatingman, line 36: http.Client can be passed by reference `*http.Client` since it is dereferenced later.

actions/monitoring/grafana.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, nil, err
+	}
+	defer resp.Body.Close()
+	body, _ := io.ReadAll(resp.Body)
+
+	if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusCreated {
+		return body, resp.Cookies(), nil
+	}
+
+	return nil, nil, fmt.Errorf("Request to %s failed with %d: %s\n", url, resp.StatusCode, string(body))
```
floatingman: You shouldn't end error strings with a newline, as it adds unnecessary space in the output.

Also, it's a bit of a Go style not to capitalize error messages.

validation/longhorn/longhorn_test.go
```diff
@@ -52,6 +57,8 @@ import (
 var (
 	LonghornEncryptionSecretBaseName       = "longhorn-crypto"
 	LonghornEncryptionStorageClassBaseName = "longhorn-crypto-global"
+	volumeBytesPrometheusQueryTemplate     = "longhorn_volume_capacity_bytes{volume=\"%s\"}"
+	numberVolumesPrometheusQuery           = "count(longhorn_volume_capacity_bytes) OR on() vector(0)"
```
floatingman: Can you add a comment on why the `OR on() vector(0)` is necessary? I think it will be helpful for future maintainers to know how to guard against the scraping interval for Prometheus counts.

validation/longhorn/schemas/pit_schemas.yaml
```diff
@@ -189,50 +189,36 @@
...
+    - action: "Query Grafana for the volume size of the newly created volume (longhorn_volume_capacity_bytes{volume=<volumeName>})"
       data: ""
-      expectedresult: "Volume metrics dashboard displays"
+      expectedresult: "Volume size shows as 1 GiB"
       position: 6
-    - action: "Open 'Longhorn - Node' dashboard"
+    - action: "Query Grafana for the new number of existing Longhorn volumes (count(longhorn_volume_capacity_bytes) OR on() vector(0))"
       data: ""
-      expectedresult: "Node metrics dashboard displays"
+      expectedresult: "Number of volumes is equal to the previously quered value + 1"
```
floatingman: quered should be queried.

actions/monitoring/grafana.go
hamistao: I believe the general standard among Go users is to pass it by value unless there is a reason to do otherwise

actions/monitoring/grafana.go
hamistao: I am not sure if we want to retry here, as failing here would probably signify a bug on the server side or maybe an ugly communication issue. But to your point, for our purposes, it may be safer to retry all the same.

validation/longhorn/longhorn_test.go
```diff
@@ -425,6 +431,82 @@ func (l *LonghornTestSuite) TestVolumeEncryption() {
...
+	_, volumeName, err := storage.CreatePVC(l.client, l.cluster.ID, charts.LonghornStorageClass)
+	require.NoError(l.T(), err)
+	l.T().Logf("Volume %s created using %s PVC", volumeName, charts.LonghornStorageClass)
+
+	// CreatePVC creates a 1GiB volume.
+	volumeSizeQuery := fmt.Sprintf(volumeBytesPrometheusQueryTemplate, strings.TrimSpace(volumeName))
+	l.T().Logf("Checking volume %s's size with Prometheus through Grafana API (%s)", volumeName, volumeSizeQuery)
+	value, err = monitoring.PrometheusQueryInGrafana(loggedClient, l.client.RancherConfig.Host, l.cluster.ID, volumeSizeQuery)
+	require.NoError(l.T(), err)
+	require.Equal(l.T(), 1*1024*1024*1024, value)
```
lscalabrini01:
```suggestion
	require.Equal(l.T(), someVariable, value)
```

Could you move the value '1*1024*1024*1024' to a variable to clarify what is this value?

hamistao: Done!

slickwarren (approved):
lgtm

please make sure this has been tested and is working properly + backports are filed appropriately

lscalabrini01 (approved):

Issue #804, opened 2026-08-06
Slice 1: Propagate system-default-registry into NeuVector chart install


floatingman:
## Summary

Propagate the `system-default-registry` Rancher setting into the NeuVector chart install payload, matching the pattern used by monitoring, logging, and other chart helpers. This is the lowest-risk change and unblocks all NeuVector airgap testing.

## Parent PRD

#803 (PRD: Make NeuVector Validation Tests Airgap-Compatible) — Slice 1

## Problem

`validation/neuvector/neuvector_hardened_test.go` builds `PayloadOpts` **without** `DefaultRegistry`:

```go
payload := actionsCharts.PayloadOpts{
    Namespace: actionsCharts.NeuVectorNamespace,
    Host:      n.client.RancherConfig.Host,
    InstallOptions: actionsCharts.InstallOptions{ Cluster: cluster, Version: ..., ProjectID: project.ID },
    K3s:      strings.Contains(n.cfg.CustomCluster.KubernetesVersion, "k3s"),
    Hardened: true,
    // ← DefaultRegistry missing
}
```

`InstallNeuVectorChart` passes `payload.DefaultRegistry` (empty string) into `NewChartInstall`, resulting in `systemDefaultRegistry: ""` in chart values. Every sibling chart helper (`ranchermonitoring.go`, `rancherlogging.go`, etc.) looks up the setting — NeuVector skips it.

Result: NeuVector pods attempt to pull from Docker Hub → `ImagePullBackOff` in airgap.

### 1. Look up `system-default-registry` in the test

In `validation/neuvector/neuvector_hardened_test.go`, before building `PayloadOpts`:

```go
registrySetting, err := n.client.Management.Setting.ByID("system-default-registry")
require.NoError(n.T(), err)
```

Set `payload.DefaultRegistry = registrySetting.Value`.

### 2. Propagate registry into NeuVector-specific chart values

NeuVector images use non-standard registry paths beyond `systemDefaultRegistry`. In `actions/charts/neuvector.go` (inside `InstallNeuVectorChart` or the values-building logic), conditionally set these when `DefaultRegistry` is non-empty:

- `registry` (global NeuVector registry prefix)
- `controller.image.repository`
- `manager.image.repository`
- `enforcer.image.repository`
- `scanner.image.repository`
- `cve.scanner.image.repository`

The exact value paths depend on the NeuVector chart version — inspect `GetChartValues` output to determine the correct structure.

### 3. Preserve non-airgap behavior

[...]

Issue #803, opened 2026-08-06
PRD: Make NeuVector Validation Tests Airgap-Compatible


floatingman:
## Problem Statement

The NeuVector validation tests (`validation/neuvector/neuvector_hardened_test.go` and `validation/charts/neuvector_test.go`) cannot run in airgap environments. Hard blockers prevent execution: a hardcoded public GitHub URL for the UI extension chart repo (resolved — see Slice 2), missing `system-default-registry` propagation into chart values, no airgap-capable pipeline, and no clear build-tag policy for airgap variants.

This means NeuVector — a security-critical component — has **zero airgap test coverage**, despite airgap being a common deployment model for security-focused customers.

## Architecture Decision: Target Existing Cluster (2026-08-10)

**Decision**: Refactor `neuvector_hardened_test.go` to target an existing pre-provisioned cluster via `rancher.clusterName` (using `clusters.NewClusterMeta`), eliminating its dependency on `qaInfraAutomation.ProvisionCustomCluster`.

**Rationale**:
- The NeuVector installation logic only requires a `cluster.ID` — provisioning is incidental to the test's actual contract
- This matches the pattern in `validation/charts/neuvector_test.go` and ~94 other validation test files that read `client.RancherConfig.ClusterName`
- Removes hard runtime dependencies on the qaInfraAutomation toolchain (ansible-core, opentofu, SSH keys, cloud credentials)
- Cluster provisioning (including airgap private-registry wiring) becomes the pipeline's sole responsibility (Slice 4)
- The `K3s` payload flag is derivable at runtime from `cluster.Provider` (`clusters.KubernetesProviderK3S`) instead of `qaInfraAutomation.customCluster.kubernetesVersion`

**Impact on Slices**:

| Slice | Impact |
|-------|--------|
| Slice 1 (#804) | Unchanged — registry propagation still needed. `K3s` derivation source changes from config to `cluster.Provider`. |
| Slice 2 (#805) | Unchanged — already implemented. UI extension targets `local`, not the downstream cluster. |
| Slice 3 (#806) | **Superseded** — test no longer provisions clusters. Private-registry wiring moves to the pipeline (Slice 4). Issue repurposed as documentation of cluster prerequisites. |
[...]

Issue #788, opened 2026-07-31
NeuVector validation tests are not airgap-compatible


floatingman:
## Summary

The NeuVector validation tests are **not airgap-compatible** in their current form. An evaluation of the two NeuVector test surfaces shows multiple hard blockers: a UI-extension chart repo hardcoded to a public GitHub URL, no propagation of the `system-default-registry` setting into chart values, an infrastructure provisioning path with no private-registry wiring, and a legacy pipeline with no airgap parameterization. This issue captures the findings and proposes a phased plan to make the NeuVector tests runnable in an airgap environment.

## Background

There are two NeuVector test surfaces:

| Test | Path | Build tag | Airgap status |
|------|------|-----------|---------------|
| Standard install | `validation/charts/neuvector_test.go` | `... && !airgap.daily` | Explicitly excluded from airgap runs by design |
| Hardened install | `validation/neuvector/neuvector_hardened_test.go` | `(validation \|\| infra.any \|\| i.cluster.any \|\| sanity \|\| pit.event) && !stress && !extended` | Not excluded, but has **no airgap adaptations** — would fail in airgap |

Supporting code:
- `validation/charts/neuvector.go` — service/web-UI verification helpers
- `actions/charts/neuvector.go` — `InstallNeuVectorChart`
- `actions/uiplugins/uiplugins.go` — `CreateExtensionsRepo`, `InstallUIPlugin`
- `interoperability/qainfraautomation/` — custom-cluster provisioning (OpenTofu + Ansible)
- `validation/neuvector/pipeline/allinone-pass/Jenkinsfile.Neuvector_allinone_pass_regression` — pipeline

### 1. UI extension repo hardcoded to public GitHub
`validation/neuvector/neuvector_hardened_test.go:42`
```go
uiPluginChartsURL = "https://github.com/rancher/ui-plugin-charts"
```
`CreateExtensionsRepo` (`actions/uiplugins/uiplugins.go`) creates a `ClusterRepo` that performs a `git clone` of this URL and waits on the `RepoDownloaded` condition. In airgap the Rancher server cannot reach `github.com`, so the repo never downloads and `SetupSuite` hangs/fails, aborting the whole suite. There is no mirror/override knob.

### 2. `system-default-registry` is never propagated
`validation/neuvector/neuvector_hardened_test.go` builds `PayloadOpts` **without** `DefaultRegistry`:
```go
payload := actionsCharts.PayloadOpts{
[...]

PR #810, opened 2026-08-07
Make NeuVector UI extension repo URL configurable for airgap

### LINKED ISSUE #805: Slice 2: Make UI extension repo URL configurable for airgap
## Summary

Replace the hardcoded `github.com` URL for the UI plugin charts repo with a config-driven value, and skip repo creation when the extension is already installed. This unblocks `SetupSuite` in airgap environments where the Rancher server cannot reach `github.com`.

## Parent PRD

#803 (PRD: Make NeuVector Validation Tests Airgap-Compatible) — Slice 2

## Problem

`validation/neuvector/neuvector_hardened_test.go:42` hardcodes:
```go
uiPluginChartsURL = "https://github.com/rancher/ui-plugin-charts"
```

`CreateExtensionsRepo` (`actions/uiplugins/uiplugins.go`) creates a `ClusterRepo` that performs a `git clone` of this URL and waits on the `RepoDownloaded` condition. In airgap the Rancher server cannot reach `github.com`, so the repo never downloads and `SetupSuite` hangs/times out, aborting the entire suite. There is no mirror/override mechanism.

### 1. Add NeuVector-specific config struct

Create a config struct (in `validation/neuvector/` or a shared config location):

```go
type NeuVectorTestConfig struct {
    UIPluginChartsURL    string `yaml:"uiPluginChartsURL"`
    UIPluginChartsBranch string `yaml:"uiPluginChartsBranch"`
    SkipUIExtension      bool   `yaml:"skipUIExtension"`
}
```

Config key: `neuvectorTest` (or nested under existing config).

### 2. Replace hardcoded constants with config values

In `SetupSuite`, load the config with sensible defaults:
- `UIPluginChartsURL` defaults to `https://github.com/rancher/ui-plugin-charts` (preserves non-airgap behavior)
- `UIPluginChartsBranch` defaults to `main`
- `SkipUIExtension` defaults to `false`

### 3. Skip repo creation when extension is pre-installed

The test already checks `IsAlreadyInstalled`. When `SkipUIExtension` is `true` OR the extension is already installed, skip the `CreateExtensionsRepo` call entirely. In airgap environments, the extension is typically pre-installed from a mirror during Rancher setup.

### 4. Support internal git mirror

When `UIPluginChartsURL` is overridden in config, pass the configured URL to `CreateExtensionsRepo`. The `ClusterRepo` will clone from the internal mirror instead of GitHub.

### 5. Test config example

```yaml
neuvectorTest:
  uiPluginChartsURL: "https://internal-git.example.com/rancher/ui-plugin-charts"
  uiPluginChartsBranch: "main"
  skipUIExtension: false
```

For airgap with pre-installed extension:
```yaml
neuvectorTest:
  skipUIExtension: true
```

## Acceptance Criteria

- [ ] `uiPluginChartsURL` is no longer hardcoded; it reads from test config with the current public URL as default
- [ ] When `skipUIExtension: true`, no `ClusterRepo` creation is attempted and no outbound `git clone` occurs
- [ ] When a mirror URL is configured, `CreateExtensionsRepo` uses that URL and `RepoDownloaded` reaches `True`
- [ ] When no config override is present, behavior is identical to today (public GitHub URL used)
- [ ] `go build -tags=validation ./validation/neuvector/...` passes

## Testing

- **Build verification**: `go build -tags=validation ./...`
- **Non-airgap regression**: Run without config override; confirm same behavior as before
- **Airgap with skip**: Set `skipUIExtension: true`; confirm no outbound git clone; confirm test proceeds past SetupSuite
- **Airgap with mirror**: Set `uiPluginChartsURL` to internal mirror; confirm repo downloads successfully

## Dependencies

- None (this slice is independent; can be done in parallel with Slice 1)

## Related

- #803 (Parent PRD)
- #788 (Source issue)
- #804 (Slice 1 — registry propagation)

floatingman:
## Summary

Closes #805 (Slice 2 of #803 — Make NeuVector Validation Tests Airgap-Compatible).

Replaces the hardcoded `github.com` URL for the UI plugin charts repo in `validation/neuvector/neuvector_hardened_test.go` with a config-driven value, and skips `ClusterRepo` creation when the extension is already installed or explicitly skipped. This unblocks `SetupSuite` in airgap environments where the Rancher server cannot reach `github.com`.

## Changes

- **`actions/neuvector/config.go`** (new): `NeuVectorTestConfig` struct loaded from the new top-level `neuvectorTest` config key, with `uiPluginChartsURL`, `uiPluginChartsBranch`, and `skipUIExtension` fields. Defaults (`https://github.com/rancher/ui-plugin-charts`, `main`) preserve existing behavior when no override is configured.
- **`validation/neuvector/neuvector_hardened_test.go`**: `SetupSuite` loads the config and applies defaults. The `IsAlreadyInstalled` check now runs first; when `skipUIExtension: true` **or** the extension is already installed, `CreateExtensionsRepo` is never called — no `ClusterRepo` creation, no outbound `git clone`. A configured mirror URL is passed through to `CreateExtensionsRepo`, which still waits on `RepoDownloaded`.
- **`validation/neuvector/README.md`**: documents the `neuvectorTest` config key with airgap examples (internal mirror and pre-installed extension).

## Config examples

Internal git mirror:
```yaml
neuvectorTest:
  uiPluginChartsURL: "https://internal-git.example.com/rancher/ui-plugin-charts"
  uiPluginChartsBranch: "main"
  skipUIExtension: false
```

Airgap with pre-installed extension:
```yaml
neuvectorTest:
  skipUIExtension: true
```

## Testing

- `go build -tags=validation ./validation/neuvector/...` ✅
- `go vet -tags=validation ./validation/neuvector/...` and `go vet ./neuvector/...` (actions module) ✅
- Runtime airgap/non-airgap verification requires live environments (per issue acceptance criteria)

## Related

- #803 (Parent PRD)
- #788 (Source issue)
- #804 (Slice 1 — registry propagation, independent)

### DIFF
--- a/actions/neuvector/config.go
+++ b/actions/neuvector/config.go
@@ -1,0 +1,21 @@
+package neuvector
+
+const (
+	// ConfigurationFileKey is the top-level key used to unmarshal neuvectorTest config from the Rancher config file.
+	ConfigurationFileKey = "neuvectorTest"
+
+	// DefaultUIPluginChartsURL is the public ui-plugin-charts repository used when no override is configured.
+	DefaultUIPluginChartsURL = "https://github.com/rancher/ui-plugin-charts"
+	// DefaultUIPluginChartsBranch is the ui-plugin-charts branch used when no override is configured.
+	DefaultUIPluginChartsBranch = "main"
+)
+
+// NeuVectorTestConfig holds configuration for NeuVector validation tests.
+type NeuVectorTestConfig struct {
+	// UIPluginChartsURL overrides the git repository used for the UI plugin charts ClusterRepo (e.g. an internal mirror in airgap environments).
+	UIPluginChartsURL string `json:"uiPluginChartsURL,omitempty" yaml:"uiPluginChartsURL,omitempty"`
+	// UIPluginChartsBranch overrides the git branch of the UI plugin charts repository.
+	UIPluginChartsBranch string `json:"uiPluginChartsBranch,omitempty" yaml:"uiPluginChartsBranch,omitempty"`
+	// SkipUIExtension skips ClusterRepo creation and UI extension install entirely (e.g. when the extension is pre-installed in airgap environments).
+	SkipUIExtension bool `json:"skipUIExtension,omitempty" yaml:"skipUIExtension,omitempty"`
+}
--- a/actions/neuvector/config.go
+++ b/actions/neuvector/config.go
@@ -16,6 +16,6 @@ type NeuVectorTestConfig struct {
 	UIPluginChartsURL string `json:"uiPluginChartsURL,omitempty" yaml:"uiPluginChartsURL,omitempty"`
 	// UIPluginChartsBranch overrides the git branch of the UI plugin charts repository.
 	UIPluginChartsBranch string `json:"uiPluginChartsBranch,omitempty" yaml:"uiPluginChartsBranch,omitempty"`
-	// SkipUIExtension skips ClusterRepo creation and UI extension install entirely (e.g. when the extension is pre-installed in airgap environments).
+	// SkipUIExtension skips ClusterRepo creation and UI extension install entirely (e.g. airgap environments where the Rancher server cannot reach github.com and no uiPluginChartsURL mirror is configured). Not required when the extension is already installed; that case is auto-detected.
 	SkipUIExtension bool `json:"skipUIExtension,omitempty" yaml:"skipUIExtension,omitempty"`
 }
--- a/validation/neuvector/README.md
+++ b/validation/neuvector/README.md
@@ -49,10 +49,19 @@ neuvectorTest:
   skipUIExtension: false
 ```
 
-Skip ClusterRepo creation and extension install entirely when the NeuVector UI extension is
-pre-installed (e.g. airgap environments where the Rancher server cannot reach github.com):
+Skip the UI extension install entirely (airgap without a `ui-plugin-charts` mirror):
+
+Use `skipUIExtension: true` when the Rancher server cannot reach `github.com` **and** no
+internal mirror is configured via `uiPluginChartsURL`. In that case the ClusterRepo would
+never sync and the install would fail, so the test skips it rather than attempting the install.
 
 ```yaml
 neuvectorTest:
   skipUIExtension: true
 ```
+
+> **Note:** This flag is **not** needed when the extension is already installed — the test
+> auto-detects an existing install via chart status and skips the install on its own.
+> With the extension skipped, the suite still installs and validates the NeuVector backend
+> chart and reaches the manager UI through the Kubernetes service proxy, so the test remains
+> meaningful without the Rancher UI extension.
--- a/validation/neuvector/README.md
+++ b/validation/neuvector/README.md
@@ -34,3 +34,25 @@ qaInfraAutomation:
 
 **Note:** Infrastructure cleanup is registered via `t.Cleanup()` inside the provisioning helpers, so it
 runs automatically when the test finishes (pass or fail).
+
+### NeuVector test options
+
+Top-level key: `neuvectorTest`. All fields are optional; without any override the test uses the public
+`https://github.com/rancher/ui-plugin-charts` repo on the `main` branch, same as before.
+
+Point the UI plugin charts ClusterRepo at an internal git mirror (e.g. airgap):
+
+```yaml
+neuvectorTest:
+  uiPluginChartsURL: "https://internal-git.example.com/rancher/ui-plugin-charts"
+  uiPluginChartsBranch: "main"
+  skipUIExtension: false
+```
+
+Skip ClusterRepo creation and extension install entirely when the NeuVector UI extension is
+pre-installed (e.g. airgap environments where the Rancher server cannot reach github.com):
+
+```yaml
+neuvectorTest:
+  skipUIExtension: true
+```
--- a/validation/neuvector/neuvector_hardened_test.go
+++ b/validation/neuvector/neuvector_hardened_test.go
@@ -6,20 +6,17 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
 	"testing"
 
 	"github.com/rancher/shepherd/clients/rancher"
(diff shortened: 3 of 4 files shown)


validation/neuvector/neuvector_hardened_test.go
```diff
@@ -71,19 +70,30 @@ func (n *NeuVectorHardenedTestSuite) SetupSuite() {
...
-	if k8sErrors.IsNotFound(err) {
-		logrus.Infof("UI plugin repo %q not found, creating it", uiPluginChartsRepoName)
-		err = uiplugins.CreateExtensionsRepo(n.client, uiPluginChartsRepoName, uiPluginChartsURL, uiPluginChartsBranch)
+	neuvectorConfig := new(neuvector.NeuVectorTestConfig)
+	operations.LoadObjectFromMap(neuvector.ConfigurationFileKey, cattleConfig, neuvectorConfig)
+	if neuvectorConfig.UIPluginChartsURL == "" {
+		neuvectorConfig.UIPluginChartsURL = neuvector.DefaultUIPluginChartsURL
+	}
+	if neuvectorConfig.UIPluginChartsBranch == "" {
+		neuvectorConfig.UIPluginChartsBranch = neuvector.DefaultUIPluginChartsBranch
```
hamistao, line 74: I would suggest following this example [here](https://github.com/hamistao/tests/blob/main/interoperability/longhorn/longhorn.go) just because it is elegant and more aligned with how it was done before here, setting a good standard for the future.

```
const (
	longhornTestConfigConfigurationFileKey = "longhorn"
	LonghornTestDefaultProject             = "longhorn-test"
	LonghornTestDefaultStorageClass        = "longhorn"
)

// GetLonghornTestConfig gets a LonghornTestConfig object using the data from the config file.
// If any data is missing default values are used.
func GetLonghornTestConfig() TestConfig {
	defer recover() // Recover from panic on LoadConfig in case no longhorn configuration is provided.

	longhornTestConfig := TestConfig{
		LonghornTestProject:      LonghornTestDefaultProject,
		LonghornTestStorageClass: LonghornTestDefaultStorageClass,
	}
	config.LoadConfig(longhornTestConfigConfigurationFileKey, &longhornTestConfig)

	return longhornTestConfig
}
```

actions/neuvector/config.go
```diff
@@ -0,0 +1,21 @@
...
+	ConfigurationFileKey = "neuvectorTest"
+
+	// DefaultUIPluginChartsURL is the public ui-plugin-charts repository used when no override is configured.
+	DefaultUIPluginChartsURL = "https://github.com/rancher/ui-plugin-charts"
+	// DefaultUIPluginChartsBranch is the ui-plugin-charts branch used when no override is configured.
+	DefaultUIPluginChartsBranch = "main"
+)
+
+// NeuVectorTestConfig holds configuration for NeuVector validation tests.
+type NeuVectorTestConfig struct {
```
hamistao, line 14: Generally it is not recommended to use  `neuvector.NeuVectorTestConfig` since the package already tells you it is neuvector so it becomes repetitive. Tex editor linters usually point that out ;)

validation/neuvector/README.md
```diff
@@ -34,3 +34,25 @@ qaInfraAutomation:
...
+  skipUIExtension: false
+```
+
+Skip ClusterRepo creation and extension install entirely when the NeuVector UI extension is
+pre-installed (e.g. airgap environments where the Rancher server cannot reach github.com):
+
+```yaml
+neuvectorTest:
+  skipUIExtension: true
+```
```
hamistao: I am confused as to the usefulness of `skipUIExtension`, as this comment says that this flag is being used to skip installation when it is already installed, but from looking at the code, when the extension is already installed the installation is skipped regardless of the flag. Could you elaborate more on this so I can understand?

validation/neuvector/neuvector_hardened_test.go
floatingman: @hamistao Thanks for the suggestion — I kept the loading inline intentionally. The same function already loads qaconfig.Config the same way via operations.LoadObjectFromMap a few lines up, so this matches the file's existing convention. The longhorn helper also needs defer recover() because config.LoadConfig panics when the key is absent; operations.LoadObjectFromMap is a no-op on a missing key (it marshals nil to "null" and unmarshalling that leaves the struct untouched), so no recover is needed here. With only one consumer today, I'd rather keep it explicit and inline; happy to extract a GetConfig helper in a follow-up if more NeuVector suites end up sharing it.

actions/neuvector/config.go
floatingman: @hamistao I checked this against the repo's lint gate — .golangci.yaml runs revive with the exported rule enabled, and it passes on this package as-is (our config doesn't enforce stuttering names; I verified by running it). There's also existing precedent in actions/ for package-qualified names, e.g., auth.AuthConfig, logging.Logging, workloads.Workloads. Since it's a style preference with no functional impact and no enforced convention behind it, I'm leaving the name as-is.

validation/neuvector/README.md
floatingman: @hamistao Good question. The IsAlreadyInstalled check covers the case where the extension was previously installed through Helm and is visible via chart status. skipUIExtension covers the case where the extension is NOT detected as installed, but the test must not attempt the install anyway — e.g., an airgap environment without a ui-plugin-charts mirror, where creating a ClusterRepo pointing at github.com would never sync, and the install would fail. The test body only exercises the NeuVector backend chart and the manager UI via service proxy, so the suite is still meaningful with the Rancher UI extension skipped entirely. Happy to clarify the README wording if you think that would help.

hamistao: Thanks for elaborating, indeed I think more detail in the README would be helpful, particularly clarifying that the test is still meaningful without the UI extension and that this example is NOT meant to illustrate a scenario where the extension is already installed but the one you just described.

slickwarren (dismissed):

validation/neuvector/README.md
floatingman: Updated the readme

hamistao (approved):

lscalabrini01 (approved):

PR #813, opened 2026-08-07
Fix kubeconfig path inconsistency

hamistao:
This has the main goal of fixing an issue where the kubeconfig path was inconsistent depending on the CWD of the running test process, as `absPath, err := filepath.Abs("kubeconfig.yaml")` in https://github.com/rancher/tests/blob/main/interoperability/qainfraautomation/provisioning.go#L1275 doesn't always align with `filepath.Join(repoPath, filepath.Dir(playbookPath), "kubeconfig.yaml")` in https://github.com/rancher/tests/blob/main/interoperability/qainfraautomation/provisioning.go#L753

To fix this, this introduces the philosophy that `clusterCfg.KubernetesVersion` should reflect the current kubeconfig location even if it wasn't set by the user initially and using it to enforce a consistent location throughout the execution of `provisionAWSStandaloneCluster` and `provisionHarvesterStandaloneCluster`.

### DIFF
--- a/interoperability/qainfraautomation/ansible/ansible.go
+++ b/interoperability/qainfraautomation/ansible/ansible.go
@@ -180,6 +180,7 @@ func (c *Client) GenerateInventoryFromNodes(clusterNodesJSON, distro, env string
 	if err != nil {
 		return "", fmt.Errorf("build inventory: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


slickwarren (approved):
lgtm -- can you link a passing run using this in our slack when you have time

lscalabrini01 (approved):

PR #816, opened 2026-08-11
Add scripts/jenkins-run.sh to trigger go-automation-freeform-job from the terminal

floatingman:
## What

Adds `scripts/jenkins-run.sh` — a single bash script to trigger Jenkins' shared `go-automation-freeform-job` from the terminal, stream the `gotestsum` console, and exit 0/1 on the build result. Lets reviewers run a new/changed test against the QA Jenkins as part of PR review without clicking through the Jenkins UI.

## Why

PR approval here requires a green Jenkins run of the new test. Today that means launching `go-automation-freeform-job` through the Jenkins UI and babysitting the console. This turns it into one terminal command whose exit code is trustworthy.

## Usage

```bash
export JENKINS_URL JENKINS_USER JENKINS_API_TOKEN
export JENKINS_JOB=rancher_qa/go-automation-freeform-job   # only if the job is foldered

./scripts/jenkins-run.sh \
  --repo https://github.com/<you>/tests \
  --branch <your-pr-branch> \
  --testcase TestFoo
```

- **Deps:** bash 4+, `curl`, `jq`. No Go, no Jenkinsfile edits, no installs.
- **Auth:** Jenkins API token via env vars; the token is never echoed (masked in `--dry-run`).

## Behavior

- CLI flags map 1:1 to the job's build parameters (`BRANCH`, `TAGS`, `TEST_PACKAGE`, `GOTEST_TESTCASE`, `CONFIG`, `TIMEOUT`, `REPO`, `QASE_TEST_RUN_ID`, `QASE_SCHEMA_PREFIX`, `ENV_VARIABLE`).
- `--config` (default `cattle-config.yaml` in cwd) sends the file **contents** as the `CONFIG` parameter (a snapshot at trigger time).
- `--testcase <regex>` emits `-run <regex>` (repo convention; values starting with `-` pass through raw).
- `--build-prefix <text>` tags the build display name `<prefix> <testcase>` for findability in the shared job history (re-asserted after the build finishes so it survives the job's own `displayName` write).
- Follows server redirects (POST-preserving), tolerates CSRF on/off, resolves the queue item, streams progressive console text.
- `--dry-run` prints the masked curl; `--detach` returns after the POST; `--help` lists everything. Exit codes: 0 success, 1 failure/cancel/rejection, 2 misuse.

## Verification

`bash -n` and `shellcheck -S warning` clean. Queue → build resolution, progressive console streaming, SUCCESS/FAILURE exit codes, `--detach`, and the display-name tag were exercised end-to-end against a mock Jenkins speaking the Remote API. Missing-env and CLI-misuse guards exit 2 with clear messages.

## Scope

One new file (`scripts/jenkins-run.sh`). No Jenkinsfile, Make, or dependency changes.

### DIFF
--- a/scripts/jenkins-run.sh
+++ b/scripts/jenkins-run.sh
@@ -1,0 +1,480 @@
+#!/usr/bin/env bash
+#
+# jenkins-run.sh — trigger Jenkins' go-automation-freeform-job from the terminal.
+#
+# Launches the freeform job with the right parameters, prints the browsable
+# build URL, streams the gotestsum console to stdout, and exits 0/1.
+#
+# Dependencies: bash 4+, curl, jq.  No Go, no Jenkinsfile edits, no installs.
+# Auth: a Jenkins API token via env vars (JENKINS_URL / JENKINS_USER / JENKINS_API_TOKEN).
+#
+# -----------------------------------------------------------------------------
+# Shell-style note (deliberate deviation — do NOT "fix" back to `set -e`):
+#
+# Other repo scripts (e.g. validation/pipeline/scripts/setup_environment.sh) use
+# `set -ex`.  This script uses `set -uo pipefail` and deliberately omits `-e`,
+# because the CSRF-crumb fetch (§3), the queue/console polling curls (§5/§6), and
+# the non-SUCCESS result handling (§7) all rely on inspecting non-zero exit
+# codes from individual commands.  With `-e` the script would abort on the first
+# transient curl failure or on the expected non-2xx crumb probe, instead of
+# looping or branching on the result.
+# -----------------------------------------------------------------------------
+
+set -uo pipefail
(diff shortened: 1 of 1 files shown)


scripts/jenkins-run.sh
```diff
@@ -0,0 +1,493 @@
+#!/usr/bin/env bash
```
lscalabrini01, line 1: Could we add a README.md file of showing how to use the script?

hamistao (approved):

scripts/jenkins-run.sh
floatingman: The help output has a good example of how to run this.

slickwarren (approved):

lscalabrini01 (approved):

PR #823, opened 2026-08-14
Update template_test.go to read template-chart.yaml instead of forked GitHub repo

markusewalker:
## Summary
Update `template_test.go` to read `template-chart.yaml` instead of sourcing the template data from a forked GitHub repository.
## Impact
This changes the test to rely on the YAML template source directly and removes the dependency on the forked repo.


susesgartner (dismissed):

validation/provisioning/k3s/defaults/template-chart.yaml
```diff
@@ -0,0 +1,59 @@
+# Required for the template test
+templateTest:
```
susesgartner: Nit, rename this to templateChart

validation/provisioning/k3s/defaults/defaults.yaml
```diff
@@ -1,11 +0,0 @@
-#Required for the template test
```
susesgartner, line 1: If there is nothing else in this file you can delete it and the defaults should focus the outer scope

validation/provisioning/rke2/defaults/defaults.yaml
```diff
@@ -1,11 +0,0 @@
-#Required for the template test
```
susesgartner, line 1: Same as above

susesgartner (dismissed):

